PR #2, opened 2016-05-26
homebrew-node: install -g npm modules with brew

ZakariaRidouh:
I've filled out the [proposal template](/proposal_template.md)


proposal_template.md
```diff
@@ -1,24 +1,15 @@
-# $FEATURE_NAME
```
MikeMcQuaid: Please create a new file rather than editing this one, thanks.

MikeMcQuaid:
This sounds like a good idea to me but it's not something that would be part of the Homebrew package manager itself. Instead this requires someone to create a third-party tap which anyone can do (but the current Homebrew maintainers will not). Thanks for the idea anyway, though!

DomT4:
The answer is probably something similar to the [brew-gem](https://github.com/sportngin/brew-gem) project if someone did want to flesh out the concept.

PR #3, opened 2016-06-02
001-implement-moveapps-and-unmoveapps.md

vitorgalvao:
Regarding

> have other advantages, such as making the app bundle both more readily recognisable by the OS and easier to get to by the user when using `Open with`.

I’ve personally been bit by that many times (and still am, occasionally) when `mpv` is updated. This is specially egregious because I’m affect by the [slow `Open with` that affects others as well](https://discussions.apple.com/thread/3865350?start=0&tstart=0).


zmwangx:
How do you handle `uninstall` in this case? Because HB keeps everything in a keg and links elsewhere, it knows exactly what to uninstall (and is smart enough to not remove files overwritten by users — which is certainly not the case for HBC). HBC doesn't, so `uninstall` with HBC is not precise and could sometimes be a pain. I for one don't want to lose the precision of `uninstall` just to make HB and HBC behave somewhat similarly (they need not).

vitorgalvao:
> How do you handle `uninstall` in this case?

That’s a good point. I actually had an idea while writing this that I abandoned, but would actually work for that case. Will updated accordingly.

> I for one don't want to lose the precision of `uninstall` just to make HB and HBC behave somewhat similarly (they need not).

You make it seem like this *addition* would take something away from you. It would not, because `linkapps` would remain.

zmwangx:
> It would not, because linkapps would remain.

Ah, okay, but precision `uninstall` is still an important point to address. IMO keeping things to kegs is part of Homebrew's philosophy, and one would need a strong reason and a bulletproof workaround to violate that. I do realize the pain caused by symlinked apps though (had a lot of experience with that due to HBC's legacy behavior).

vitorgalvao:
> Ah, okay, but precision `uninstall` is still an important point to address.

Just tried it, and `uninstall` did not remove the link. Seems like `unlinkapps` is still needed.

> a bulletproof workaround

I believe I have one. Writing it up now. Basically we would create the symlink in reverse (from `/Applications` to the Cellar). With that, at `uninstall` we need only verify where the Cellar’s app symlink points to, and remove that.

vitorgalvao:
Updated the proposal (also added `unmoveapps`).

zmwangx:
> Basically we would create the symlink in reverse

That's a reasonable solution (actually I did exactly that with some HBC-installed apps before), but you don't know if user has overwritten the application, and when you delete you could delete stuff that's not managed by you. I'm not sure if that's desirable.

vitorgalvao:
> but you don't know if user has overwritten the application, and when you delete you could delete stuff that's not managed by you.

At some point the responsibility must shift from the software to the user, and I believe that to be the point, here. Say I installed [`youtube-dl`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/youtube-dl.rb) and at some point ran `youtube-dl --update`. Now I may have an issue, and it won’t be accurately reported since the version reported by HB is incorrect. This was the user’s fault: they should have left the updating to HB (that is the goal of the software) and decided to override that. To be clear, I do also believe that is their right as user and that HB should not interfere with that freedom. However, the user is the responsible one for their break of the normal usage of the software.

Same thing with `moveapps`. Like `linkapps`, this is a command that must be *explicitly given*. The consequences of such a command are pretty explicit and not hard to understand. If you know you ran that command and decide to overwrite the app without proper sanitation (`unmoveapps`), at this points it was the user’s responsibility.

With all of this in mind, even the chances of an overwrite of this kind of something that could not be easily recuperated are so slim, I don’t see it as a valid argument to not implement the feature. Not only is it a very simple outcome to understand, it is entirely optional.

vitorgalvao:
> HBC doesn't, so `uninstall` with HBC is not precise and could sometimes be a pain.

Small parenthesis there. You just gave me one idea to be safer about that. We could add HBC-installed app bundles to a new group (shared by the user) and at `uninstall` we’d only delete stuff if in that group (and hence we’d be sure it belonged to us).

However, this “dumbness” of HBC is many times a feature. Things like `zap` and `uninstall` on a `pkg` are real assets to clean things up, even if you installed them manually, and can in theory do better than a manual cleanup.

MikeMcQuaid:
Great proposal @vitorgalvao and well written! The main problem we have with our application bundles is that they are often no relocatable either due to hardcoded paths, relative paths or missing dependencies inside the application bundle itself. Any thoughts for a workaround for that?

UniqMartin:
Thank you for the work you put into writing a proposal! I'm excited to see the first proposal of its kind!

> The main problem we have with our application bundles is that they are often no relocatable either due to hardcoded paths, relative paths or missing dependencies inside the application bundle itself. Any thoughts for a workaround for that?

I believe this is the most significant technical difficulty that will need to be addressed to make the proposal feasible. And how much of a problem this really is hugely depends on the formula. App bundles are really supposed to be independent entities, but those built by Homebrew as part of a larger formula build process are very rarely are completely independent.

> Just tried it, and `uninstall` did not remove the link. Seems like `unlinkapps` is still needed.

This closely ties into the above paragraph about app bundles not being completely independent from the files living in the corresponding formula prefix. I don't think there are major technical challenges to the task of keeping track of where an app bundle that now lives in `/Applications` comes from. What I do worry most about is the behavior on `upgrade`. I'd really like to see this fleshed out in the proposal, i.e. what the expected behavior is for apps that have been moved with `moveapps` and whose origin formula is now being upgraded. For `[un]linkapps` the situation is *a lot* simpler because the provided apps of a formula rarely changes, so in most situations the previously created symbolic link just continues to work.

**TL;DR** What I think needs to be addressed in this proposal in more detail is how `[un]moveapps` interacts with `uninstall`, `upgrade`, and `cleanup` and what the expected behavior is.

---

*Workflow-related side note:* Among all the repositories across the Homebrew organization, I think this is one where new commits instead of one amended commit should be the norm for incremental changes to the proposal, so that people who have kept track of the discussion have a chance to understand how raised issues have been addressed. (At least I find it hard to pick out the differences from the original proposal I saw and that has subsequently been updated, so I needed to re-read it completely.) Not sure though, if everyone feels like me about this … we're still only getting started with this repository.

MikeMcQuaid:
> Among all the repositories across the Homebrew organization, I think this is one where new commits instead of one amended commit should be the norm for incremental changes to the proposal, so that people who have kept track of the discussion have a chance to understand how raised issues have been addressed.

@UniqMartin Agreed 👍 I was thinking of suggesting we do this for all non-formula changes at this point as it's much easier to follow non-trivial review.

vitorgalvao:
@MikeMcQuaid

> they are often no relocatable

I thought that *might* be the case, but since the ones I know of in HB also exist as casks in HBC, I was hoping it wouldn’t actually *be* the case.

> due to hardcoded paths, relative paths or missing dependencies inside the application bundle itself.

Are these by choice (i.e. HB builds them that way, but they could easily be self-contained) or limitations (even if we wanted to, there’s no way we could make them self-contained with a reasonable amount of effort)?

If the former, may I get an insight as to what the reasoning is (perhaps updating the components without needing to rebuild the bundle)? Note I’m not saying at all this is the wrong decision, just that I need to understand it to come up with a reasonable solution.

If the latter, may I get two or three examples of formulas where this happens, and why that is the case?

---

@UniqMartin 

> what the expected behavior is for apps that have been moved with `moveapps` and whose origin formula is now being upgraded

Of the top of my head (i.e. still need to think of flaws) I’m thinking `upgrade` would check for the symlink, and if it exists, do `unmoveapps {{formula}}` and `moveapps {{formula}}` after the upgrade.

> For `[un]linkapps` the situation is a *lot* simpler because the provided apps of a formula rarely changes, so in most situations the previously created symbolic link just continues to work.

How so? Aren’t the app bundles left inside a versioned Cellar directory, meaning it changes with upgrades? Note I usually get rid of old versions on every upgrade, so there might be some behaviour here I’m not considering.

---

Thank you both for the kind words. Also, making new commits makes perfect sense, will do future ones like that.

zmwangx:
@vitorgalvao 

> How so? Aren’t the app bundles left inside a versioned Cellar directory, meaning it changes with upgrades?

I believe `brew linkapps` links from the opt prefix, i.e., `$(brew --prefix)/opt/foo` instead of from a versioned keg, so it rarely changes.

MikeMcQuaid:
> Are these by choice (i.e. HB builds them that way, but they could easily be self-contained) or limitations (even if we wanted to, there’s no way we could make them self-contained with a reasonable amount of effort)?

Limitations of the upstream build system, usually. If you're building your app outside Xcode with other Homebrew dependencies it's actually vaguely tricky to ensure you've done this correctly (fun aside: I created https://cmake.org/cmake/help/v3.0/module/DeployQt4.html for CMake to do this for Qt applications in the past).

> If the former, may I get an insight as to what the reasoning is (perhaps updating the components without needing to rebuild the bundle)? Note I’m not saying at all this is the wrong decision, just that I need to understand it to come up with a reasonable solution.

It's definitely never the former. I think it would be really reasonable and actively desirable to adopt Homebrew Casks approach of moving in the medium-long term and detect application bundles that can't be moved somehow (`otool -L` on their binaries/libraries will probably be enough) and say that Homebrew will not build/install/include those application bundles in future.

> If the latter, may I get two or three examples of formulas where this happens, and why that is the case?

I can go looking for these but another maintainer may beat me to it.

vitorgalvao:
> It's definitely never the former. I think it would be really reasonable and actively desirable to adopt Homebrew Casks approach of moving in the medium-long term and detect application bundles that can't be moved somehow (`otool -L` on their binaries/libraries will probably be enough) and say that Homebrew will not build/install/include those application bundles in future.

I’m trying to make this proposal take the least amount of work and work as best as possible with what is already in place, but seeing you find this desirable is a huge plus.

Perhaps the next step would then be to [figure out what can be deduplicated](https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-223563011), removing entirely the formulas that will exist solely on HBC’s side, and figure out how to proceed based on what remains. Do you agree? If yes, should I add that task to the proposal?

From a quick search, it doesn’t seem like there’s any flag to `grep` for or similar to find all formulas that provide app bundles, is there?

UniqMartin:
> From a quick search, it doesn’t seem like there’s any flag to `grep` for or similar to find all formulas that provide app bundles, is there?

Correct. The suggestion to use `linkapps` after installing a formula that provides app bundles entirely depends on a check whether app bundles can be found in the formula prefix. There's no metadata in the formula itself that would allow identifying those formulae with 100% certainty.

You can probably get an incomplete listing by grepping for `.app` since many build systems place the app bundles in `bin` and we tend to move them somewhere else (usually the formula prefix or `libexec`), thus you will find formulae that have code to move the app bundles.

vitorgalvao:
On a preliminary run, these seem to be the formulas from `homebrew-core` that can install app bundles (feel free to edit the list): 

+ [apcupsd](https://github.com/Homebrew/homebrew-core/blob/master/Formula/apcupsd.rb)
+ [avidemux](https://github.com/Homebrew/homebrew-core/blob/master/Formula/avidemux.rb)
+ [cppcheck](https://github.com/Homebrew/homebrew-core/blob/master/Formula/cppcheck.rb)
+ [djview4](https://github.com/Homebrew/homebrew-core/blob/master/Formula/djview4.rb)
+ [emacs](https://github.com/Homebrew/homebrew-core/blob/master/Formula/emacs.rb)
+ [gammaray](https://github.com/Homebrew/homebrew-core/blob/master/Formula/gammaray.rb)
+ [graphviz](https://github.com/Homebrew/homebrew-core/blob/master/Formula/graphviz.rb)
+ [kdiff3](https://github.com/Homebrew/homebrew-core/blob/master/Formula/kdiff3.rb)
+ [macvim](https://github.com/Homebrew/homebrew-core/blob/master/Formula/macvim.rb)
+ [mpv](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mpv.rb)
+ [pinentry-mac](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pinentry-mac.rb)
+ [plan9port](https://github.com/Homebrew/homebrew-core/blob/master/Formula/plan9port.rb)
+ [puddletag](https://github.com/Homebrew/homebrew-core/blob/master/Formula/puddletag.rb)
+ [qcachegrind](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qcachegrind.rb)
+ [qjackctl](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qjackctl.rb)
+ [qt](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt.rb)
+ [qt5](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt5.rb)
+ [valkyrie](https://github.com/Homebrew/homebrew-core/blob/master/Formula/valkyrie.rb)
+ [wireshark](https://github.com/Homebrew/homebrew-core/blob/master/Formula/wireshark.rb)

---

Some of those are not present in HBC (will need to evaluate them further to see if they can be added). A few questions:

From [above](https://github.com/Homebrew/brew-evolution/pull/3#issuecomment-224329361):

> Perhaps the next step would then be to [figure out what can be deduplicated](https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-223563011), removing entirely the formulas that will exist solely on HBC’s side, and figure out how to proceed based on what remains. Do you agree? If yes, should I add that task to the proposal?

Also, should I add these as an issue on [homebrew-core](https://github.com/Homebrew/homebrew-core) in the same vein of [HBC’s deduplication issue](https://github.com/caskroom/homebrew-cask/issues/15603) so the community can discuss which ones can safely be removed in favour of casks, and also improve the list?

MikeMcQuaid:
> Also, should I add these as an issue on homebrew-core in the same vein of HBC’s deduplication issue so the community can discuss which ones can safely be removed in favour of casks, and also improve the list?

@vitorgalvao Makes sense to me. @AnastasiaSulyagina may also have thoughts. We can also consider deprioritising those that only do so optionally for now.

MikeMcQuaid:
Note that anything build with Qt (e.g. prefixed with a `q` and `gammaray`) should be able to use `macdeployqt` to build something relocatable.

AnastasiaSulyagina:
I have everything that installs .app as cask by default if formulas are present in both brew-core and cask (seems logical and consistent for users). Some formulas like wireshark have non-gui behavior as well so I would leave them for now.

MikeMcQuaid:
I _think_ we have some sort of agreement here that we will not implement this proposal but instead start to gradually deprecate `linkapps`, right?

UniqMartin:
> I _think_ we have some sort of agreement here that we will not implement this proposal but instead start to gradually deprecate `linkapps`, right?

I'm in favor of this resolution.

Despite all its flaws, `linkapps` works somewhat well with our current system because it leaves all the relevant bits where they were installed to. Building app bundles that are truly relocatable and independent of all other formulae floating around in a Homebrew installation—something a reliably working `moveapps` would depend on—feels like it is out of scope for Homebrew, where significant portions of our formulae are tied to the prefix/cellar they were built for.

MikeMcQuaid:
Closing but thanks for the discussion everyone 👍

vitorgalvao:
Sounds good. Apologies for the somewhat late reply, I’m currently abroad and don’t have my laptop or a regular internet connection.

PR #5, opened 2016-07-16
Command to test all formulas

RSully:


UniqMartin:
Thanks for taking the time to write a proposal! I think your use case is already covered by what is available today. I'm not sure if that's obvious or not, but you can pass multiple formula names to `brew test`, e.g. `brew test ack boost` will test both `ack` and `boost`.

Using this knowledge, you can leverage `brew list` (that lists all installed formulae) and feed its output to `brew test`, i.e.:

```sh
brew test $(brew list)
```

This will often return with an error, but in many cases that's because there are still formulae without a test and testing formulae without a test is considered an error. Does this help with your use case? Are there any points from your proposal that are not addressed by this?

RSully:
I didn't know that; really helpful. In that case, perhaps `brew test` could add a flag to not treat missing tests as errors.

When running that command I am a bit surprised how many tests actually fail on my system. I'll have to look into this before saying much more.

vitorgalvao:
> In that case, perhaps `brew test` could add a flag to not treat missing tests as errors.

Although I see where you’re coming from, I’d actually consider a feature not having said flag. If there’s no way to bypass the failure, it may incentivise the person to work on a test (hence benefiting the formula for everyone).

MikeMcQuaid:
> Using this knowledge, you can leverage brew list (that lists all installed formulae) and feed its output to brew test

Yeh, this is what I do too. We tend to try and favour commands that can be composed using typical Unix style like this.

> Although I see where you’re coming from, I’d actually consider a feature not having said flag. If there’s no way to bypass the failure, it may incentivise the person to work on a test (hence benefiting the formula for everyone).

I agree 👍

DomT4:
Also agree with @vitorgalvao on `brew test` failing hard on formulae missing tests. Formulae missing tests is a consistent problem when it comes to breaking changes slipping through CI, and 100% test coverage is a desire _(Even if not always a practical one, depending on how the packaged software is intended to behave)_.

RSully:
I noticed I can't ctrl+c out of `brew test (brew list)`, it kills the child process but then goes on the test the next one. Even hammering on it doesn't kill it. 

There are a lot of formulas that fail, even after reinstalling, for example homebrew/homebrew-core#3201. Should I be creating an issue for each of these (or pull requests for the easy fixes when possible)?

RSully:
I do agree with many of the thoughts in this thread, though. The only downside is that it is dependent on the shell, so if you wanted to execute it programmatically you either have to do it within a shell process or execute list, then test, separate. Not really a problem, just worth a small mention.

MikeMcQuaid:
Passing on this for now due to discussion above but thanks for the PR!

PR #6, opened 2016-07-27
003 Install bottles of arbitrary versions

sjackman:
Install bottles without having access to the formula used to create them.

### DIFF
--- a/003-install-bottles-of-arbitrary-versions.md
+++ b/003-install-bottles-of-arbitrary-versions.md
@@ -12,7 +12,7 @@ Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's
 Store the formula used to build the bottle inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
 
 ## Detailed design
-When installing a keg from source, store the formula inside the keg in a file named `NAME/VERSION/.bottle/NAME.rb`. When building a bottle from that keg, store the formula used to build the keg inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the bottle block or `sha256` lines.
+When installing a keg from source, store the names of the dependencies and the versions used to build the keg inside `INSTALL_RECEIPT.json`. When building a bottle from that keg, store the formula used to build the keg inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the bottle block or `sha256` lines.
 
 ## Alternatives considered
 1. Store the git SHA1 of the revision of the formula used to build the bottle in the bottle so that the formula can be later extracted with `git show`.
--- a/003-install-bottles-of-arbitrary-versions.md
+++ b/003-install-bottles-of-arbitrary-versions.md
@@ -6,13 +6,15 @@ Install bottles without having access to the formula used to create them.
 ## Motivation
 The motivation for this feature is to install a collection of bottles of arbitrary versions to repeat a data analysis pipeline. It is one step toward the loftier goal of being able to recreate the environment of a previous installation of Homebrew. See the original feature request and discussion at [Homebrew/homebrew-science#1191](https://github.com/Homebrew/homebrew-science/issues/1191).
 
-## Proposed solution
 Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's dependencies nor its `post_install` are currently stored in the bottle, and the information in the current formula may have changed since the bottle was built. It would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
 
-The formula use to build the bottle could be stored inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
+## Proposed solution
+Store the formula used to build the bottle inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
 
 ## Detailed design
-When building a bottle, store the formula inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the bottle block or `sha256` lines.
+When installing a keg from source, store the formula inside the keg in a file named `NAME/VERSION/.bottle/NAME.rb`. When building a bottle from that keg, store the formula used to build the keg inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the bottle block or `sha256` lines.
 
 ## Alternatives considered
-Find the formula in the `git log` as `git versions` did.
+1. Store the git SHA1 of the revision of the formula used to build the bottle in the bottle so that the formula can be later extracted with `git show`.
+2. Store the `git hash-object` SHA1 of the formula used to build the bottle in the bottle so that the formula can later be extracted with `git cat-file blob`.
+3. Find the formula in the `git log` as `brew versions` did.
--- a/003-install-bottles-of-arbitrary-versions.md
+++ b/003-install-bottles-of-arbitrary-versions.md
@@ -1,0 +1,21 @@
+# Install bottles of arbitrary versions
+
+## Introduction
+I'd like to be able to install bottles of arbitrary versions of a formula.
+
+## Motivation
+The motivation for this feature is to install specified versions of software to repeat a data analysis pipeline, an oft-requested feature of Homebrew-Science, which could be happily stored in a Brewfile. See Homebrew/homebrew-science#1191
+
+## Proposed solution
+Pouring the bottle (extracting the tarball) is easy enough. The bottle's dependencies and its `post_install` are the needed missing info.
+
+One simple way of storing this info is to store the Ruby formula that generated the bottle alongside the bottle. To that end, when uploading bottles to Bintray, I'd also like to upload the Ruby formula that generated the bottle.
+
+That Ruby formula could alternatively be stored inside the bottle, with the caveat that it would be necessary to remove the bottle `sha256`, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode. As a bonus, it would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
+
+## Detailed design
+When building a bottle, store the Ruby formula inside the bottle in a file named `NAME/VERSION/NAME.rb` after removing the bottle `sha256` lines.
+
+## Alternatives considered
+1. Store the Ruby formula inside the bottle (without the bottle `sha256` lines)
+2. Store the Ruby formula alongside the bottle on Bintray
--- a/003-install-bottles-of-arbitrary-versions.md
+++ b/003-install-bottles-of-arbitrary-versions.md
@@ -4,7 +4,7 @@
 Install bottles without having access to the formula used to create them.
 
 ## Motivation
-The motivation for this feature is to install a collection of bottles of arbitrary versions to repeat a data analysis pipeline. See [Homebrew/homebrew-science#1191](https://github.com/Homebrew/homebrew-science/issues/1191).
+The motivation for this feature is to install a collection of bottles of arbitrary versions to repeat a data analysis pipeline. It is one step toward the loftier goal of being able to recreate the environment of a previous installation of Homebrew. See the original feature request and discussion at [Homebrew/homebrew-science#1191](https://github.com/Homebrew/homebrew-science/issues/1191).
 
 ## Proposed solution
 Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's dependencies nor its `post_install` are currently stored in the bottle, and the information in the current formula may have changed since the bottle was built. It would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
@@ -12,7 +12,7 @@ Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's
 The formula use to build the bottle could be stored inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
 
 ## Detailed design
-When building a bottle, store the formula inside the bottle in a file named `NAME/VERSION/NAME.rb` after removing the bottle block or `sha256` lines.
+When building a bottle, store the formula inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the bottle block or `sha256` lines.
 
 ## Alternatives considered
 Find the formula in the `git log` as `git versions` did.
--- a/003-install-bottles-of-arbitrary-versions.md
+++ b/003-install-bottles-of-arbitrary-versions.md
@@ -1,21 +1,18 @@
 # Install bottles of arbitrary versions
 
 ## Introduction
-I'd like to be able to install bottles of arbitrary versions of a formula.
+Install bottles without having access to the formula used to create them.
 
 ## Motivation
-The motivation for this feature is to install specified versions of software to repeat a data analysis pipeline, an oft-requested feature of Homebrew-Science, which could be happily stored in a Brewfile. See Homebrew/homebrew-science#1191
+The motivation for this feature is to install a collection of bottles of arbitrary versions to repeat a data analysis pipeline. See [Homebrew/homebrew-science#1191](https://github.com/Homebrew/homebrew-science/issues/1191).
 
 ## Proposed solution
-Pouring the bottle (extracting the tarball) is easy enough. The bottle's dependencies and its `post_install` are the needed missing info.
+Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's dependencies nor its `post_install` are currently stored in the bottle, and the information in the current formula may have changed since the bottle was built. It would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
 
-One simple way of storing this info is to store the Ruby formula that generated the bottle alongside the bottle. To that end, when uploading bottles to Bintray, I'd also like to upload the Ruby formula that generated the bottle.
-
-That Ruby formula could alternatively be stored inside the bottle, with the caveat that it would be necessary to remove the bottle `sha256`, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode. As a bonus, it would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
+The formula use to build the bottle could be stored inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
 
 ## Detailed design
-When building a bottle, store the Ruby formula inside the bottle in a file named `NAME/VERSION/NAME.rb` after removing the bottle `sha256` lines.
+When building a bottle, store the formula inside the bottle in a file named `NAME/VERSION/NAME.rb` after removing the bottle block or `sha256` lines.
 
 ## Alternatives considered
-1. Store the Ruby formula inside the bottle (without the bottle `sha256` lines)
-2. Store the Ruby formula alongside the bottle on Bintray
+Find the formula in the `git log` as `git versions` did.


003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Motivation
+The motivation for this feature is to install specified versions of software to repeat a data analysis pipeline, an oft-requested feature of Homebrew-Science, which could be happily stored in a Brewfile. See Homebrew/homebrew-science#1191
+
+## Proposed solution
+Pouring the bottle (extracting the tarball) is easy enough. The bottle's dependencies and its `post_install` are the needed missing info.
+
+One simple way of storing this info is to store the Ruby formula that generated the bottle alongside the bottle. To that end, when uploading bottles to Bintray, I'd also like to upload the Ruby formula that generated the bottle.
+
+That Ruby formula could alternatively be stored inside the bottle, with the caveat that it would be necessary to remove the bottle `sha256`, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode. As a bonus, it would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
```
MikeMcQuaid: I'd say you should remove the entirely `bottle` block. Should be relatively easily done with the existing regexes.

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,21 @@
+# Install bottles of arbitrary versions
+
+## Introduction
+I'd like to be able to install bottles of arbitrary versions of a formula.
+
+## Motivation
+The motivation for this feature is to install specified versions of software to repeat a data analysis pipeline, an oft-requested feature of Homebrew-Science, which could be happily stored in a Brewfile. See Homebrew/homebrew-science#1191
```
MikeMcQuaid: Can you make `Homebrew/homebrew-science#1191` a full link? I also don't think this motivation is matched by the feature; you'll only get a single specified version that's the same. Also, storing this in a `Brewfile` would be another big implementation detail that would need to be discussed separately (and I'm not convinced is a good idea). Could you trim this down so it's just about the motivation for installing bottles?

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,21 @@
+# Install bottles of arbitrary versions
+
+## Introduction
+I'd like to be able to install bottles of arbitrary versions of a formula.
```
MikeMcQuaid: How about: `install bottles without having access to the formula used to create them`? That seems more accurate as what you propose is already possible.

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Motivation
+The motivation for this feature is to install specified versions of software to repeat a data analysis pipeline, an oft-requested feature of Homebrew-Science, which could be happily stored in a Brewfile. See Homebrew/homebrew-science#1191
+
+## Proposed solution
+Pouring the bottle (extracting the tarball) is easy enough. The bottle's dependencies and its `post_install` are the needed missing info.
+
+One simple way of storing this info is to store the Ruby formula that generated the bottle alongside the bottle. To that end, when uploading bottles to Bintray, I'd also like to upload the Ruby formula that generated the bottle.
```
MikeMcQuaid: I'd pick one method; personally I think the latter is a stronger case.

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,21 @@
+# Install bottles of arbitrary versions
+
+## Introduction
+I'd like to be able to install bottles of arbitrary versions of a formula.
+
+## Motivation
+The motivation for this feature is to install specified versions of software to repeat a data analysis pipeline, an oft-requested feature of Homebrew-Science, which could be happily stored in a Brewfile. See Homebrew/homebrew-science#1191
+
+## Proposed solution
+Pouring the bottle (extracting the tarball) is easy enough. The bottle's dependencies and its `post_install` are the needed missing info.
```
MikeMcQuaid: Can you elaborate on why they are missing info?

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,21 @@
...
+
+One simple way of storing this info is to store the Ruby formula that generated the bottle alongside the bottle. To that end, when uploading bottles to Bintray, I'd also like to upload the Ruby formula that generated the bottle.
+
+That Ruby formula could alternatively be stored inside the bottle, with the caveat that it would be necessary to remove the bottle `sha256`, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode. As a bonus, it would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
+
+## Detailed design
+When building a bottle, store the Ruby formula inside the bottle in a file named `NAME/VERSION/NAME.rb` after removing the bottle `sha256` lines.
```
MikeMcQuaid: I'd use somewhere in https://github.com/Homebrew/brew/blob/0be204aab5878284d8466b141a82d56a166abe7b/Library/Homebrew/formula.rb#L705-L711 instead; perhaps with just `formula.rb` so you don't need to parse the name from the bottle.

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Alternatives considered
```
MikeMcQuaid: These should be alternatives to implementing this proposal rather than the alternatives for implementing it.

003-install-bottles-of-arbitrary-versions.md
sjackman: I think it would be good for the bottle `revision` of the formula to reflect the bottle revision of the bottle's file name. It would default to`0` if deleted, without logic to fix it up.

MikeMcQuaid: The bottle's `revision` does not affect anything in formula installation, only the bottle filename.

003-install-bottles-of-arbitrary-versions.md
UniqMartin: I think this is far from being sufficient, thus currently doesn't fulfill the stated goal of the proposal. To achieve some degree of reproducibility, you would need to keep track of a lot more information. I've detailed that in https://github.com/Homebrew/brew/issues/589#issuecomment-235434219 in the original issue.

003-install-bottles-of-arbitrary-versions.md
sjackman: I have an inclination toward retaining the bottle block and revision, but I don't feel terribly strongly about it.

003-install-bottles-of-arbitrary-versions.md
sjackman: > I also don't think this motivation is matched by the feature; you'll only get a single specified version that's the same.

A list of formula names and versions or a directory of bottle files could be used to install a collection of software.

> Could you trim this down so it's just about the motivation for installing bottles?

Sure.

003-install-bottles-of-arbitrary-versions.md
sjackman: Are you suggesting `NAME/VERSION/.bottle/formula.rb`?

The name of the formula file is usually `NAME.rb`, so I stuck with that convention. I'm fine with either.

I put it in the same directory as `INSTALL_RECEIPT.json`, which is `NAME/VERSION/`. What's the effect/advantage of putting it in `NAME/VERSION/.bottle/`?

sjackman:
I've modified my proposal based on your feedback. Thanks for your helpful comments.

003-install-bottles-of-arbitrary-versions.md
MikeMcQuaid: I think it'd be easier implementation to just remove it; it's metadata that's not used.

003-install-bottles-of-arbitrary-versions.md
MikeMcQuaid: > Are you suggesting NAME/VERSION/.bottle/formula.rb?

I was referring to just inside the tarball but presumably you mean inside the Cellar in which case: yes.



> What's the effect/advantage of putting it in NAME/VERSION/.bottle/?

It's metadata that's only needed for the bottle and can be removed when the bottle is installed.



> I think this is far from being sufficient, thus currently doesn't fulfill the stated goal of the proposal. To achieve some degree of reproducibility, you would need to keep track of a lot more information.

@UniqMartin I don't agree. Currently you can install any bottle and it'll just assume the formula is the same as the current one. This would be an improvement on that. I don't think reproducibility is really a sensible requirement at this stage and it massively balloons the scope.

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,18 @@
...
+## Proposed solution
+Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's dependencies nor its `post_install` are currently stored in the bottle, and the information in the current formula may have changed since the bottle was built. It would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
+
+The formula use to build the bottle could be stored inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
+
+## Detailed design
+When building a bottle, store the formula inside the bottle in a file named `NAME/VERSION/NAME.rb` after removing the bottle block or `sha256` lines.
+
+## Alternatives considered
+Find the formula in the `git log` as `git versions` did.
```
scpeters: I think @UniqMartin's comment in https://github.com/Homebrew/brew/issues/589#issuecomment-235434219 could be listed here as an alternative. The approach is more invasive, but more thorough in respecting the goal of reproducibility.

I also had a comment about a tool that helps with extracting the HEAD commit from the taps ( https://github.com/Homebrew/brew/issues/589#issuecomment-235438453 ).

sjackman: Being able to reproduce bottles is a separate goal. The goal of this PR is to be able to install preexisting bottles.

scpeters: If the old bottles are cached on bintray, this method should allow the old bottles to be installed, as long as brew doesn't try to auto-update, since everything is purposefully back-dated.

UniqMartin: > Being able to reproduce bottles is a separate goal. The goal of this PR is to be able to install preexisting bottles.

I think you're misunderstanding me. What I've been talking about in the first half of the referenced comment is exactly that: Accurately re-create an environment that was previously created by installing a bunch of bottles. Because the behavior of what is going on during bottle installation and `post_install` also depends on other factors outside of the formula file, keeping around only the formula file that was used during bottling (by embedding it in the bottle) is a first step, but will have its limitations.

MikeMcQuaid: > is a first step, but will have its limitations.

That's fine for now; everything needs to begin with a first step.

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,18 @@
+# Install bottles of arbitrary versions
+
+## Introduction
+Install bottles without having access to the formula used to create them.
+
+## Motivation
+The motivation for this feature is to install a collection of bottles of arbitrary versions to repeat a data analysis pipeline. See [Homebrew/homebrew-science#1191](https://github.com/Homebrew/homebrew-science/issues/1191).
```
MikeMcQuaid: After reading this link: it's maybe worth removing it as, as @UniqMartin has pointed out, this won't get people what they want there and to do that will require enough work to be essentially impossible. It's maybe worth just rewording to say it's the first step required to get there?

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Proposed solution
+Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's dependencies nor its `post_install` are currently stored in the bottle, and the information in the current formula may have changed since the bottle was built. It would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
+
+The formula use to build the bottle could be stored inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
+
+## Detailed design
+When building a bottle, store the formula inside the bottle in a file named `NAME/VERSION/NAME.rb` after removing the bottle block or `sha256` lines.
```
MikeMcQuaid: Missing the `.bottle` here.

sjackman: What's the advantage of naming the file `NAME/VERSION/.bottle/NAME.rb` vs `NAME/VERSION/NAME.rb`? `INSTALL_RECEIPT.json` is named `NAME/VERSION/INSTALL_RECEIPT.json`.

MikeMcQuaid: This is bottle-specific metadata that we'll discard after installation rather than install. It's easier to have `.bottle` be the only location that stores things that aren't simply `tar xf`d (e.g. currently `.bottle/etc` and `.bottle/var`).

sjackman: Thanks for the explanation, Mike.

sjackman:
I've revised my proposal based on your feedback. Thanks again for your helpful comments.

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Proposed solution
+Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's dependencies nor its `post_install` are currently stored in the bottle, and the information in the current formula may have changed since the bottle was built. It would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
+
+The formula use to build the bottle could be stored inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
+
+## Detailed design
+When building a bottle, store the formula inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the bottle block or `sha256` lines.
```
sjackman: Is `NAME/VERSION/.bottle/NAME.rb` installed in the keg? After installation of the bottle, a few commands need the formula for the keg. For example, `brew missing` needs to know its dependencies, and `brew postinstall` needs its `post_install` block.

UniqMartin: > Is `NAME/VERSION/.bottle/NAME.rb` installed in the keg?

No, the stuff in `NAME/VERSION/.bottle` is used when pouring the bottle, but then removed afterwards (see [these lines in `FormulaInstaller#pour`](https://github.com/Homebrew/brew/blob/97b6a3069ea909d2dc6987f49f3e47eeda7054b1/Library/Homebrew/formula_installer.rb#L748-L752)).

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,18 @@
...
+## Introduction
+Install bottles without having access to the formula used to create them.
+
+## Motivation
+The motivation for this feature is to install a collection of bottles of arbitrary versions to repeat a data analysis pipeline. It is one step toward the loftier goal of being able to recreate the environment of a previous installation of Homebrew. See the original feature request and discussion at [Homebrew/homebrew-science#1191](https://github.com/Homebrew/homebrew-science/issues/1191).
+
+## Proposed solution
+Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's dependencies nor its `post_install` are currently stored in the bottle, and the information in the current formula may have changed since the bottle was built. It would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
+
+The formula use to build the bottle could be stored inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
```
bfontaine: Small typo:

> The formula use**d** to build

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Detailed design
+When building a bottle, store the formula inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the bottle block or `sha256` lines.
+
+## Alternatives considered
+Find the formula in the `git log` as `git versions` did.
```
bfontaine: Did you mean `brew versions` instead?

003-install-bottles-of-arbitrary-versions.md
sjackman: Shall we move the formula to a different location, or shall we make an exception and install `NAME/VERSION/.bottle/NAME.rb`?

003-install-bottles-of-arbitrary-versions.md
sjackman: Fixed and reworded. Thanks.

003-install-bottles-of-arbitrary-versions.md
sjackman: Woops. Fixed. Thanks.

003-install-bottles-of-arbitrary-versions.md
MikeMcQuaid: I don't think we should install it; I think that's creeping towards @UniqMartin's suggestions where we'll need to record a lot more information. Then again, in that case we're basically only using it for `post_install`.

003-install-bottles-of-arbitrary-versions.md
```diff
@@ -0,0 +1,20 @@
...
+
+Pouring the bottle (extracting the tarball) is easy enough. Neither the bottle's dependencies nor its `post_install` are currently stored in the bottle, and the information in the current formula may have changed since the bottle was built. It would be a nice feature of the bottle file format if the bottle itself contained all of the necessary information to install the bottle.
+
+## Proposed solution
+Store the formula used to build the bottle inside the bottle. It would be necessary to remove the `bottle` section or the `sha256` lines, as it wouldn't be possible to store the `sha256` of the bottle inside the bottle, lest the universe implode.
+
+## Detailed design
+When installing a keg from source, store the formula inside the keg in a file named `NAME/VERSION/.bottle/NAME.rb`. When building a bottle from that keg, store the formula used to build the keg inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the bottle block or `sha256` lines.
```
sjackman: The formula used to build the keg needs to be stored somewhere between building the keg and building the bottle. Does the above make sense?

sjackman:
Revised both `Detailed design` and `Alternatives considered`. Moved a paragraph from `Proposed solution` to `Motivation`.

003-install-bottles-of-arbitrary-versions.md
sjackman: I'd prefer to install it, primarily to enable `brew missing`. Once a bunch of bottles are installed, it'd be nice to check that their dependencies are satisfied using `brew missing`.

MikeMcQuaid: Sorry to keep yoyoing (thinking is hard) but I think if we're going to install it instead we should put those dependencies in the `INSTALL_RECEIPT`.

sjackman: No worries. In that case let's add the dependencies to `INSTALL_RECEIPT.json` when the keg is built from source, and store the formula used to build the bottle in the bottle at `NAME/VERSION/.bottle/NAME.rb`. Do we want to store the `post_install` block anywhere in the keg to support running `brew postinstall NAME` after the bottle has been installed?

sjackman:
Revised `Detailed design` to store the dependencies (and their versions?) in `INSTALL_RECEIPT.json`.

MikeMcQuaid:
Looks good to me. Any final thoughts?

sjackman:
It won't be possible to run `brew postinstall` after installing the keg unless we store the `post_install` block somewhere inside the keg. Is that a feature we care about?

MikeMcQuaid:
> It won't be possible to run brew postinstall after installing the keg unless we install the post_install block somewhere inside the keg. Is that a feature we care about?

@sjackman I think we could either hold off deleting it until after it's been post_installed or just leave it in `.bottle` indefinitely.

sjackman:
I'm fine with leaving it in `.bottle`.

MikeMcQuaid:
Thanks @sjackman and everyone!

sjackman:
Thanks for the helpful reviews, all! I'm excited to have the first merged proposal. The process worked well.

sjackman:
Who's interested in implementing it? I can tackle it, but I won't start the project for a week or two. Go right ahead if anyone is keener than that.

MikeMcQuaid:
I doubt I'll get to it before then. I'd suggest two PRs: one to install the formula, one to add to the INSTALL_RECEIPT.

sjackman:
Okay. I'll hoping to hack on it while I'm in Michigan giving a workshop on genome assembly with @tseemann.

PR #7, opened 2016-08-05
Add an install option to keep debugging symbols

justinclift:


DomT4:
See previous: https://github.com/Homebrew/legacy-homebrew/pull/42053

justinclift:
[Ouch](https://github.com/Homebrew/legacy-homebrew/pull/42053#issuecomment-126942899).  Looks like that was a solid first attempt.

Wonder if there's a different way to do things?  Or perhaps this PR would be better targeted towards minor changes to brew, so that "debug friendly formulas" are feasible in taps after all?

DomT4:
It's a difficult thing to get done, not only on our side but because OS X doesn't really handle retaining symbols in a particularly sensible way as mentioned by megahall in that thread.

That PR was definitely a good start & I applaud OP for the willingness to take a swing at it for sure. The baseline for us ideally is something like `option :universal` with the command-line flag `--universal`. Users see almost none of the magic that happens around making `:universal` _"just work"_.

The problem increasingly with the proposed `--dsym` feature was that as much as not it required in-formula changes, and not particularly insignificant ones at that, and in-formula changes essentially bind every future contributor to those formulae to understanding & re-tweaking as necessary the `dysm` steps.

That & the need to build inside the formula prefix rather than a temporary buildpath were the two biggest hurdles we couldn't really work out how to jump over at the time.

justinclift:
With the need to build inside the formula prefix, a few potential reasons that is bad come to mind:

* Incomplete builds (eg in-progress ones) might lead to weird mis-detection problems in running applications
* Could failed builds have a negative effect?

Those are the two main negatives springing to mind from that aspect of things.  Other there others too (specifically for the in-prefix build)?

MikeMcQuaid:
Yeh, passing on this due to the reasons @DomT4 stated above; we'd need a system that was more-or-less automatic and that doesn't seem to be supported by OS X, unfortunately.

PR #8, opened 2016-08-13
004 List packages by install date (or recent packages)

ajlowndes:
[content of md file repeated here:]

# LIST PACKAGES BY INSTALL DATE
## Introduction
List "recent" packages homebrew has installed.

## Motivation
Installed a bunch of tools which had a bunch of dependencies. Used third-party taps. I have now finished with these tools and the dependencies and want to remove them, but I have no idea what their names were (and they are days ago in my bash history). What did I install that day? 

## Proposed solution
A new brew command, e.g. `brew list --date DDMMYY` or `brew list recent X` where X is the most recent X number of packages.

## Detailed design
If you run `brew info [package]` on an installed package, brew tells you the date it was poured. So this info is logged, we just need to search by this field.

## Alternatives considered
Can't think of any.

### DIFF
--- a/004-list-packages-by-install-date.md
+++ b/004-list-packages-by-install-date.md
@@ -1,0 +1,15 @@
+# LIST PACKAGES BY INSTALL DATE
+## Introduction
+List "recent" packages homebrew has installed.
+
+## Motivation
+Installed a bunch of tools which had a bunch of dependencies. Used third-party taps. I have now finished with these tools and the dependencies and want to remove them, but I have no idea what their names were (and they are days ago in my bash history). What did I install that day? 
+
+## Proposed solution
+A new brew command, e.g. `brew list --date DDMMYY` or `brew list recent X` where X is the most recent X number of packages.
+
+## Detailed design
+If you run `brew info [package]` on an installed package, brew tells you the date it was poured. So this info is logged, we just need to search by this field.
+
+## Alternatives considered
+Can't think of any.


ajlowndes:
should really have called them **formulae** instead of **packages**. Sorry!

MikeMcQuaid:
`brew ls -t` does this already. It's maybe worth mentioning in the manpage that anything arguments to `brew ls` are passed through to `ls` which does the actual listing.

ajlowndes:
great - so `brew list -ltT` gives me the list of formulae I installed/updated and the date/time I installed/updated them. Thanks!

MikeMcQuaid:
@ajlowndes How do you think we could have documented this better so you would have discovered it?

ajlowndes:
I guess as you said before - I would have looked at the ls man page if `brew list --help` had mentioned that `brew list` passes everything to `ls` so you can look there for further functionality. cheers for the swift response!

bfontaine:
> Installed a bunch of tools which had a bunch of dependencies. Used third-party taps. I have now finished with these tools and the dependencies and want to remove them

You may be interested by [`brew-bundle`](https://github.com/Homebrew/homebrew-bundle). Make a bundle of what you have installed on your machine and when you want to remove unneeded formulae run  `brew bundle cleanup`.

ajlowndes:
cool, thank you @bfontaine

PR #9, opened 2016-08-14
004 Store homebrew settings in configuration files

jcsilkey:


004-read-homebrew-options-from-config-file.md
```diff
@@ -0,0 +1,34 @@
...
+current environmental variable method.
+
+Currently, the only way to set Homebrew options is via environmental variable. There is
+no way to share settings between users or system jobs setting them individually.
+
+## Detailed design
+A prototype implementation can be found at
+https://github.com/jcsilkey/microbrew/tree/feature/homebrew-config-file.
+
+1. Upon execution, `brew` would search /etc/homebrew/homebrew.json and
```
MikeMcQuaid: Definitely won't be looking in `/etc` for this given we don't encourage users to set things in these system-wide locations.

MikeMcQuaid:
Personally I'm 👎 on this; I don't think we have enough that warrants this level of configuration beyond just using environment variables as-is. If we do start using configuration files I'd like to see use have a simple file that can be sourced as a Bash script e.g. `VAR=value` which is significantly more human readable/writable than alternatives like JSON or YAML.

jcsilkey:
I don't know if I can persuade you as to whether this is valuable or not, but as to JSON vs YAML vs INI vs just straight sourcing, I chose JSON simply because:

1. I don't like to straight source unknown input.
2. I wanted to be able to whitelist valid options.
3. It was easy to write a quick and dirty ruby one-liner to parse a JSON string for demonstration purposes.
4. While its probably overkill for a simple use case like this, I don't think JSON format is *that* unreadable:

`{
  "HOMEBREW_NO_EMOJI": "1",
  "HOMEBREW_CACHE": "/usr/local/var/homebrew/cache"
}`

That said, I'm not married to using JSON as the file format.

004-read-homebrew-options-from-config-file.md
jcsilkey: Removed.

004-read-homebrew-options-from-config-file.md
```diff
@@ -0,0 +1,33 @@
...
+1. Upon execution, `brew` would search $HOMEBREW_PREFIX/etc/homebrew/homebrew.json
+for a configuration file and load it if it exists.
+2. `brew` would then search for $HOME/.homebrew/homebrew.json and load it if found.
+3. Any options previously set via environmental variables are ignored, allowing users
+to continue setting options via the current process.
+4. The configuration loader utilizes a whitelist to only load specific settings
+from the configuration file.
+
+## Alternatives considered
+1. Adding environmental variables in /etc/profile or /etc/bashrc.
```
tdsmith: Can you describe some of the advantages this approach would provide over setting environment variables in shell init scripts? I don't think they're coming through for me yet.

tdsmith:
Some secondary bikeshedding (I think getting a sense of the advantages of this approach is more important): the Python distutils-sig recently spilled several railroad cars of ink choosing on a format for packaging metadata and someone eventually chose toml; it quacks like a well-defined .ini file. It's non-executable, friendlier than json (comments and trailing commas!), and simpler than yaml. Probably the upshot is that there are many, many options; they're probably all fine, though some of them come at the cost of vendoring a parser.

jcsilkey:
1. It's allows for a consistent configuration across the system for all users and processes. I think it's safe to say most Macs are not used as multi-user machines, but for those that are I think this is beneficial. I originally coded up my prototype because I was experimenting with Puppet and homebrew was throwing errors because HOMEBREW_CACHE wasn't set. While I eventually figured out that Puppet allows you to set environmental variables when executing scripts, it got me thinking that it would be better to have it set globally and not to worry about it if I ever ran into the problem again. I suppose an alternative approach also could have been to patch homebrew to still work if the cache wasn't configured.

2. I feel like it's a more common practice to configure programs via a configuration file than setting environmental variables. That may just be my opinion though. 

3. As proposed and in my prototype code, it supplements but doesn't replace using environmental variables. It's 100% opt in. 

4. This was also partly inspired by other changes I'm experimenting with but I'm not yet ready to share, for which it would be extremely helpful if not necessary. Not that that helps make my case here. :)

jcsilkey:
@tdsmith I probably would have gone with yaml since it's simpler than JSON, but I didn't because I couldn't see anywhere else it was used in homebrew while there was at least one instance of JSON (install receipts). Given that it's just a list of key: value pairs and not a complicated structure, I personally don't feel like there's a great deal of difference between yaml and and an inifile (or almost any other format), except yaml doesn't require any vendored code.

004-read-homebrew-options-from-config-file.md
jcsilkey: @tdsmith I think I responded in the wrong place to your comment. Hopefully I addressed your question ok?

MikeMcQuaid:
> I don't like to straight source unknown input.

That's not necessarily implied by using a Bash initialisation file. You could read in something, split on `=`, discard every key not starting with `HOMEBREW_`, discard every invalid value.

> Probably the upshot is that there are many, many options; they're probably all fine, though some of them come at the cost of vendoring a parser.

Agreed. I'm a hard 👎 on introducing another parser or format. Homebrew currently uses JSON, git's configuration and environment variables and only the latter for user configuration. If (and it's a big if) we implement this then it should lean as closely as possible to mirroring existing environment variables and being very human-writable; JSON is just too easy to write in an invalid way.

jcsilkey:
@MikeMcQuaid I think I read it too literally when you said "source as a bash script". While I have my personal preference (YAML, FWIW - it's well supported by ruby and I think it's pretty easy to read and write, especially if you're not dealing with nested structures), that's not a hill I want to die on.

So where does that leave this?

MikeMcQuaid:
I'd like to wait a bit for other @Homebrew/maintainers to chip in.

ilovezfs:
I'm told Linuxbrew is a thing, but dare I say the word "plist"?

ilovezfs:
I'm also wondering how all of this would interact with `brew bundle`.

MikeMcQuaid:
> I'm also wondering how all of this would interact with brew bundle.

@ilovezfs I would see this as being `source`d in some form from Bash so it would affect all commands.

ilovezfs:
@MikeMcQuaid I just mean that inevitably people will want to be able to set these parameters in the dump file itself instead of having to pair it with a second file, and then we have two formats for the same info, or annoying UX where the bundle is split over two files. :see_no_evil:

Anyway, so far the status quo of just standard environment variables and nothing else seems better than anything being discussed here.

jcsilkey:
@ilovezfs As I currently see this, it would interact identically to how the current environmental variables do. If you look at https://github.com/jcsilkey/microbrew/blob/feature/homebrew-config-file/Library/Homebrew/config.sh and https://github.com/jcsilkey/microbrew/blob/feature/homebrew-config-file/Library/Homebrew/read_config_file.rb in my fork (admittedly this is not production ready code), it parses the config file and exports the options as environmental variables in the homebrew process.

UniqMartin:
I'm slightly in favor of the general idea, but I'm currently quite opposed to using JSON, YAML, or any other configuration file format that cannot be trivially handled from Bash code. The file will need to be processed on every Homebrew invocation and spawning a separate Ruby process just to do that feels prohibitively expensive.

> If we do start using configuration files I'd like to see use have a simple file that can be sourced as a Bash script e.g. `VAR=value` which is significantly more human readable/writable than alternatives like JSON or YAML.

Agreed. The only thing I'm not entirely sure about is what level of validation would be useful. I can easily see a too strict validation to prematurely reduce the usefulness of such a configuration file, e.g. if we only allowed for `VAR=value` where `value` is always interpreted as a fixed string, then we would be blocking useful things like:

```shell
HOMEBREW_CACHE="$HOMEBREW_PREFIX/.brew-cache"
HOMEBREW_LOGS="$HOMEBREW_PREFIX/.brew-logs"
```

> I'm told Linuxbrew is a thing, but dare I say the word "plist"?

Of all the possible file format options this seems to be the worst. What would be the benefit of using a `.plist` over any of the other discussed file formats?

ilovezfs:
>Of all the possible file format options this seems to be the worst. What would be the benefit of using a .plist over any of the other discussed file formats?

Not reinventing the wheel on the platform.

Anyway, I just do this in .bashrc:
```
source ~/.bashrc_homebrew
```
Problem is already solved a while ago at least for me.

MikeMcQuaid:
> Anyway, I just do this in .bashrc:
> source ~/.bashrc_homebrew
> Problem is already solved a while ago at least for me.

Yeh, the more I think about this the more it feels like any solution will be badly reimplementing what you can already do with shell configuration files and adds more code we need to maintain and fix bugs in for minimal benefit. Passing on this for now but thanks for the write-up.

UniqMartin:
@jcsilkey Thanks for coming up with the proposal and a demo implementation! Overall I have to agree that this is probably too much of a niche use case to be supported natively by us.

FWIW, I think this would be useful (admittedly only for a minority of users like myself) if such a configuration file was loaded from a location relative to the Homebrew prefix/repository. I happen to frequently switch between multiple local Homebrew installations that I wish to use with different settings. This currently entails purging my environment of any `HOMEBREW_*` variables and then setting them anew for the given Homebrew installation, followed by adjusting `PATH`. If this feature was implemented, this would simplify this workflow to adjusting `PATH` or even allow for directly invoking `brew` of a different Homebrew installation by supplying its full path.

MikeMcQuaid:
@UniqMartin https://github.com/direnv/direnv may be of interest.

jcsilkey:
Thank you to everybody for your feedback. Since this is closed, I've decided to convert my demo to a wrapper script that reads a config file before invoking brew (https://github.com/jcsilkey/microbrew). Feedback and contributions are of course always welcome. :)

Before signing off, I have some other questions, is the google discussion group the appropriate forum for asking them? Related to that, would that have been a better avenue to have raised this suggestion prior to opening a PR here?

Thank you again for the feedback and for everything else you do to keep homebrew going for the rest of us!

MikeMcQuaid:
> Thank you to everybody for your feedback. Since this is closed, I've decided to convert my demo to a wrapper script that reads a config file before invoking brew (https://github.com/jcsilkey/microbrew). Feedback and contributions are of course always welcome. :)

@jcsilkey Cool!

> Before signing off, I have some other questions, is the google discussion group the appropriate forum for asking them? Related to that, would that have been a better avenue to have raised this suggestion prior to opening a PR here?

Nah, your PR here was great.



> Thank you again for the feedback and for everything else you do to keep homebrew going for the rest of us!

Thanks for the kind words and for your PR1

PR #10, opened 2016-08-23
004-better-tap-migrations.md

adidalal:
As requested by @MikeMcQuaid in Slack

### DIFF
--- a/004-better-tap-migrations.md
+++ b/004-better-tap-migrations.md
@@ -1,14 +1,14 @@
 # A more flexible `tap_migrations.json`
 
 ## Introduction
-Allow for formulas to be pointed to differently-named formula/casks in `tap_migrations.json`
+Allow for formulas to be pointed to differently-named Casks in `tap_migrations.json`
 
 ## Motivation
-Came up while starting work on deprecating `homebrew-binary`. Would also be generally useful to deal with formula renames.
-A specific context: ngrok2 can be :skull: from homebrew-binary, but the cask is titled ngrok.
+Came up while starting work on deprecating `homebrew-binary`.
+A specific context: `ngrok2` can be :skull: from homebrew-binary, but the cask is titled `ngrok`.
 
 ## Proposed solution
-Optionally, append the (different) Formula/Cask name. Otherwise, assume identical (current behavior)
+Optionally, append the (different) Cask name. Otherwise, assume identical naming (current behavior). This change should be limited to dealing with Formula -> Cask migrations.
 
 ## Detailed design
 ```json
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Makes sense to me. Will give people a little bit to chime in with other thoughts.

UniqMartin:
I think the extension to `tap_migrations.json` itself is fine, but I wonder about the finer details of the implementation. I'm not sure how much of an issue this is for casks, as I'm not a Cask user, but for formulae this would need to tap into most of the formulae rename logic because we would need to preserve the old name in the cellar and in `opt` for various reasons (one being dynamic linkage to stuff at the old location) and move everything to the new name.

To me, this ultimately raises the question whether we should merge `tap_migrations.json` and `formulae_renames.json` into a single file if we start mixing the two. Both are actually about moving formulae (and more recently casks), it's just inter- and intra-tap moves and obviously the code behind the renames is a bit more involved because so much is tight to a stable formula name as an identifier (file system paths, dependencies, etc.).

MikeMcQuaid:
@UniqMartin Perhaps, for now, we could allow this just for Cask as it's much simpler and barf if it's a formula?

UniqMartin:
> Perhaps, for now, we could allow this just for Cask as it's much simpler and barf if it's a formula?

That would be fine by me. I just wanted to point out the potential complications and things worth considering if we think about extending this mechanism to formulae.

In any case, I'm happy about this proposal because it made me aware that the two mentioned migration mechanisms are conceptually very similar and could be unified, even though they differ quite a bit in their implementation. Thanks for that, @adidalal!

adidalal:
Another, (related) issue that just came up was Cask renames - might it make sense to consider a `migrations.json` which handles the superset of both cases, for both Formulas and Casks?

ilovezfs:
If cask renames within a given tap worked (apparently that's just never been worked on), isn't this achieved with the existing infrastructure by just doing a migration and then a rename? I guess simplifying things with a unified implementation is aesthetically more pleasing, but it seems to me there's actually nothing needed here other than basic intra-tap cask rename support and the functional goal is achieved.

adidalal:
@ilovezfs That sounds reasonable to me, no sense overcomplicating things (and certainly, anything is better than the current non-exsistent rename support)

MikeMcQuaid:
@adidalal Can you modify this just to be scoped to migrating things to Homebrew/cask for now and then I'll merge it? Thanks!

adidalal:
Updated, thanks!

MikeMcQuaid:
Thanks @adidalal!

PR #11, opened 2016-08-25
Rearrange files and some other tweaks

UniqMartin:
Put proposals into a separate directory to avoid cluttering up the repository root once the number of proposals becomes bigger. Also make the proposal number four digits because this reads nicer.

Make sure proposal numbers start at one and future proposals are numbered such that we end up with a list of integers without any gaps. (This may in the future require some pre-merge adjustments to the PR branch once discussion has ended, but this feels like a minor burden and will only happen if there are two simultaneous proposal PRs of which one gets merged earlier than the other, thus infrequently.) Explain this naming convention in the template.

Add a “Status: Accepted” line to the only currently accepted proposal, so that this can be changed to “Implemented” (with a link to the implementing PR) once it's done.


MikeMcQuaid:
> Put proposals into a separate directory to avoid cluttering up the repository root once the number of proposals becomes bigger.

I'm wondering if this puts too much Git knowledge on contributors when they inevitably create it in the root.

> Make sure proposal numbers start at one

I think they should correspond to the PR number.

Otherwise 👍

UniqMartin:
> I'm wondering if this puts too much Git knowledge on contributors when they inevitably create it in the root.

I'm not sure I understand this comment. Maybe the proposals were up until now created in the root because there was no `proposals` directory, no other proposal to look at, and no instructions to do something else. It shouldn't be a hassle to create a new file in the `proposals` directory once it's there, or am I missing something?

> I think they should correspond to the PR number.

You mean the PR number in this repository? This will give us a fairly sparsely populated space of proposal numbers, as the majority will probably be rejected or there are PRs (like this one), that don't actually contain a proposal. Personally I'm not a big fan of this and think what I'm proposing here would be nicer. (In any case, the current proposal number is neither, but a seemingly random “003” when looking at the list without considering the history.)

MikeMcQuaid:
> It shouldn't be a hassle to create a new file in the proposals directory once it's there, or am I missing something?

It shouldn't be but it might be. Maybe worth adding a single-click link as part of this to create a new file there like https://github.com/Homebrew/brew-evolution/new/master/.proposals/ ?

> You mean the PR number in this repository? This will give us a fairly sparsely populated space of proposal numbers, as the majority will probably be rejected or there are PRs (like this one), that don't actually contain a proposal. Personally I'm not a big fan of this and think what I'm proposing here would be nicer. (In any case, the current proposal number is neither, but a seemingly random “003” when looking at the list without considering the history.)

I think that's fine. A proposal is a proposal regardless of acceptance.

UniqMartin:
> It shouldn't be but it might be. Maybe worth adding a single-click link as part of this to create a new file there like https://github.com/Homebrew/brew-evolution/new/master/.proposals/ ?

I don't know. Is there a way to modify the above URL such that the file name and file contents are already populated? I have to admit that I was thinking more in terms of someone doing all this stuff locally. Doing this directly from the GitHub site still feels strange to me, but I can definitely see how that might be appealing to many people and also lowers the barrier to entry.

> I think that's fine. A proposal is a proposal regardless of acceptance.

Yes, that's true. But shouldn't we be merging them then and just adding a status of “Rejected” to them? (Not saying this is a good idea, just thinking out loud.) No matter what we ultimately end up with, we could make things a bit easier for interested parties by tagging PRs in here as “accepted proposal”, “rejected proposal”, or “non-proposal”. How does that sound?

One concern I have with asking people to number their proposal by PR number is that people won't know the PR number in advance and it's a bit harder to determine the next number that will likely be assigned to their PR than looking up the last used number from a directory listing. Do you have an idea how this could be addressed?

MikeMcQuaid:
> Is there a way to modify the above URL such that the file name and file contents are already populated?

Yep e.g. https://github.com/Homebrew/brew-evolution/new/master/.proposals/?filename=0000-proposal.md



> Yes, that's true. But shouldn't we be merging them then and just adding a status of “Rejected” to them? (Not saying this is a good idea, just thinking out loud.) No matter what we ultimately end up with, we could make things a bit easier for interested parties by tagging PRs in here as “accepted proposal”, “rejected proposal”, or “non-proposal”. How does that sound?

Definitely agreed with labeling PRs 👍. Think it's nicer to avoid merging but don't feel strongly. Don't personally feel the non-continuous numbering is a problem but happy for others to disagree.

> One concern I have with asking people to number their proposal by PR number is that people won't know the PR number in advance and it's a bit harder to determine the next number that will likely be assigned to their PR than looking up the last used number from a directory listing. Do you have an idea how this could be addressed?

I do not. This is a good point. Perhaps it's worth avoiding numbering at all or doing it pre-merge?

zmwangx:
Any update here? We already have a fully implemented proposal (`005-mkdir-with-block-create-intermediaries.md`), keeping it in the top level without any sort of distinction is confusing and counter-productive.

Either we

> Add a “Status: Accepted” line to the only currently accepted proposal, so that this can be changed to “Implemented” (with a link to the implementing PR) once it's done.

or create a `completed`/`implemented`/`archive` or whatever directory and move proposals there once implemented. We don't necessarily need a `proposals` directory, I think it's fine to keep accepted ones in the top level.

MikeMcQuaid:
@zmwangx I'm happy with any of those options. I don't think @UniqMartin is around currently. Could you open a new PR?

UniqMartin:
As pointed out by Mike, I unfortunately won't be able to revive this PR in the immediate future. @zmwangx I'm happy for you to give this another shot in a PR of your own. Just a few thoughts:

- Put all proposals into a single directory. That way they won't have to be moved around when their status changes. Aside from uncluttering the repository root when the number of proposals continues to increase, it also makes it easier to link to them from elsewhere (and avoid that the link gets broken subsequently).

- Put the status of the proposal (accepted, implemented, …) into a header at the top of the proposal. It could also be useful to link the PR that lead to the proposal from there.

- Maintain a simplified version of <https://apple.github.io/swift-evolution/> (of course adapted to our needs) at the bottom of the README or some other suitable place if we need to have an overview of all declined/accepted proposals and their implementation status.

PR #12, opened 2016-09-09
Avoid system Python

tdsmith:
This proposal suggests that Homebrew abandon its long-standing position
of favoring system Python for applications whenever possible.

I'm hoping to collect feedback from Homebrew maintainers and the Python
community on this PR; let me know if you feel that's inappropriate.

### DIFF
--- a/XX-avoid-system-python.md
+++ b/XX-avoid-system-python.md
@@ -1,0 +1,46 @@
+# Avoid system Python
+
+Tim D. Smith, Sep 9 2016
+
+## Introduction
+
+This proposal suggests that Homebrew abandon its long-standing position of favoring system Python for applications whenever possible.
+
+## Motivation
+
+The primary reason to prefer Homebrew's Python over system Python is that system Python links the `ssl` module included in the Python stdlib against the end-of-lifed 0.9.8 branch of OpenSSL.
+
+OpenSSL 0.9.8 does not support TLS 1.1 or 1.2. Despite attacks, there are safe modes of operation for TLS 1.0, but TLS 1.0 is nonetheless being phased out. For example, the PCI (Payment Card Industry) standards have set a [sunset date](https://blog.pcisecuritystandards.org/migrating-from-ssl-and-early-tls) for TLS 1.0 and vendors including CDN Fastly [are responding apace](https://www.fastly.com/blog/update-our-tls-10-and-11-deprecation-plan) by progressively disabling TLS 1.0 traffic.
+
+Applications including mercurial have begun to complain if TLS 1.1+ connections cannot be established. Homebrew responded by installing mercurial [against a Homebrew python by default](https://github.com/Homebrew/homebrew-core/issues/3541); as we move boldly into the future, more applications are likely to exhibit this behavior.
+
+## Mitigations
+
+The popular HTTP client library `requests` will prefer PyOpenSSL to the stdlib ssl module if it is available. The `cryptography` package, upon which PyOpenSSL depends, may be built against a Homebrew OpenSSL. This means that applications which rely on `requests` for all network traffic are not affected if the formula installs the necessary additional dependencies.
+
+Alas: not all TLS traffic is HTTP, and not all applications handle HTTP traffic with `requests`.
+
+## Proposed solution
+
+* Begin using `depends_on :python` with reckless abandon for applications that use python at runtime. For most users, most of the time, this will resolve to a dependency on Homebrew python. Users building from source will, subject to resolution of https://github.com/Homebrew/brew/issues/712, build using the first python in $PATH instead.
+* Add audit checks that look for suspicious Python shebangs or virtualenvs.
+* Continue to allow implicit dependencies on system Python for library bindings, since those are unlikely to interact with the ssl module.
+* Continue to allow implicit build dependencies on system Python, unless network traffic is involved.
+
+Concerns with this approach are:
+
+* If you don't want to install Homebrew's Python, it is inconvenient to globally opt out.
+  * A possible mitigation would be to introduce yet another environment variable to force formulas with a :python dependency to build from source.
+  * Another possible mitigation is the [`--with-custom-python`](https://github.com/Homebrew/homebrew-core/pull/3548/files#diff-79aaf9c9ba469f17b9f16f996d7e6290R17) switch deployed for Mercurial, which has the desired effect but introduces class-level conditional dependencies, which don't behave well.
+* Homebrew uses one fewer system dependency.
+* ?
+
+
+## Alternatives
+
+* Do nothing: since the majority of Homebrew users use the most recently released version of OS X, Homebrew could elect to solve this issue by inaction if the forthcoming Sierra release linked the system python against a more recent TLS library. Unfortunately, the betas released to date continue to build the ssl module against OpenSSL 0.9.8.
+* Use `depends_on "python"` instead. This will always resolve to a dependency on Homebrew python. This is more reliable and more obvious; the bottling behavior for :python is surprising if you haven't spent much time with it. It'll also annoy people that don't want to use Homebrew's python.
+
+## Recommendation
+
+Adopt the proposed solution with zero or more mitigations.
--- a/006-avoid-system-python.md
+++ b/006-avoid-system-python.md
@@ -33,7 +33,6 @@ Concerns with this approach are:
   * A possible mitigation would be to introduce yet another environment variable to force formulas with a :python dependency to build from source.
   * Another possible mitigation is the [`--with-custom-python`](https://github.com/Homebrew/homebrew-core/pull/3548/files#diff-79aaf9c9ba469f17b9f16f996d7e6290R17) switch deployed for Mercurial, which has the desired effect but introduces class-level conditional dependencies, which don't behave well.
 * Homebrew uses one fewer system dependency.
-* ?
 
 
 ## Alternatives
@@ -44,3 +43,7 @@ Concerns with this approach are:
 ## Recommendation
 
 Adopt the proposed solution with zero or more mitigations.
+
+## Resolution
+
+The proposed solution is adopted without mitigations.


XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+
+Tim D. Smith, Sep 9 2016
+
+## Introduction
+
+This proposal suggests that Homebrew abandon its long-standing position of favoring system Python for applications whenever possible.
+
+## Motivation
+
+The primary reason to prefer Homebrew's Python over system Python is that system Python links the `ssl` module included in the Python stdlib against the end-of-lifed 0.9.8 branch of OpenSSL.
```
MikeMcQuaid: Also worth noting: Apple have told people this is deprecated and they should move away from it. Shame they haven't 😀

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Mitigations
+
+The popular HTTP client library `requests` will prefer PyOpenSSL to the stdlib ssl module if it is available. The `cryptography` package, upon which PyOpenSSL depends, may be built against a Homebrew OpenSSL. This means that applications which rely on `requests` for all network traffic are not affected if the formula installs the necessary additional dependencies.
+
+Alas: not all TLS traffic is HTTP, and not all applications handle HTTP traffic with `requests`.
+
+## Proposed solution
+
+* Begin using `depends_on :python` with reckless abandon for applications that use python at runtime. For most users, most of the time, this will resolve to a dependency on Homebrew python. Users building from source will, subject to resolution of https://github.com/Homebrew/brew/issues/712, build using the first python in $PATH instead.
```
MikeMcQuaid: I think it may be worth just making `depends_on :python` always use the Homebrew Python and providing another option for people to specify a custom Python if they really want one. The current situation causes weird failures with `pyenv` or Anaconda Python's in the PATH and I'm not sure users really care what Python is being used if it's up-to-date for anything except e.g. Vim (where we already do this)

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+* Add audit checks that look for suspicious Python shebangs or virtualenvs.
+* Continue to allow implicit dependencies on system Python for library bindings, since those are unlikely to interact with the ssl module.
```
MikeMcQuaid: 👍

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+## Mitigations
+
+The popular HTTP client library `requests` will prefer PyOpenSSL to the stdlib ssl module if it is available. The `cryptography` package, upon which PyOpenSSL depends, may be built against a Homebrew OpenSSL. This means that applications which rely on `requests` for all network traffic are not affected if the formula installs the necessary additional dependencies.
+
+Alas: not all TLS traffic is HTTP, and not all applications handle HTTP traffic with `requests`.
+
+## Proposed solution
+
+* Begin using `depends_on :python` with reckless abandon for applications that use python at runtime. For most users, most of the time, this will resolve to a dependency on Homebrew python. Users building from source will, subject to resolution of https://github.com/Homebrew/brew/issues/712, build using the first python in $PATH instead.
+* Add audit checks that look for suspicious Python shebangs or virtualenvs.
```
MikeMcQuaid: 👍

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+* Continue to allow implicit dependencies on system Python for library bindings, since those are unlikely to interact with the ssl module.
+* Continue to allow implicit build dependencies on system Python, unless network traffic is involved.
```
MikeMcQuaid: 👍

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+
+Concerns with this approach are:
+
+* If you don't want to install Homebrew's Python, it is inconvenient to globally opt out.
```
MikeMcQuaid: The same argument could be made for OpenSSL and I think the same logic increasingly applies to both.

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+  * A possible mitigation would be to introduce yet another environment variable to force formulas with a :python dependency to build from source.
+  * Another possible mitigation is the [`--with-custom-python`](https://github.com/Homebrew/homebrew-core/pull/3548/files#diff-79aaf9c9ba469f17b9f16f996d7e6290R17) switch deployed for Mercurial, which has the desired effect but introduces class-level conditional dependencies, which don't behave well.
```
MikeMcQuaid: We could add logic to the requirement for this somehow that does something smarter.

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+* Homebrew uses one fewer system dependency.
```
MikeMcQuaid: This is sad but in this case it's increasingly looking like Python is implicitly deprecated by Apple too.

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+* ?
+
+
+## Alternatives
+
+* Do nothing: since the majority of Homebrew users use the most recently released version of OS X, Homebrew could elect to solve this issue by inaction if the forthcoming Sierra release linked the system python against a more recent TLS library. Unfortunately, the betas released to date continue to build the ssl module against OpenSSL 0.9.8.
+* Use `depends_on "python"` instead. This will always resolve to a dependency on Homebrew python. This is more reliable and more obvious; the bottling behavior for :python is surprising if you haven't spent much time with it. It'll also annoy people that don't want to use Homebrew's python.
```
MikeMcQuaid: I'm a fairly big 👍 on this.

MikeMcQuaid:
Great proposal. A bunch of comments agreed but I agree we should do something.

XX-avoid-system-python.md
tdsmith: Anaconda is difficult to support because it has its own system package management philosophy. I think pyenv generally works fine and that it's helpful to support it because I tell people to use it if they want older branches of Python 3. :)

MikeMcQuaid: I'm 🆒 with supporting pyenv I'm just not sure it's necessary to have e.g. networking using applications use it by default when building from source.

XX-avoid-system-python.md
```diff
@@ -0,0 +1,46 @@
...
+* If you don't want to install Homebrew's Python, it is inconvenient to globally opt out.
+  * A possible mitigation would be to introduce yet another environment variable to force formulas with a :python dependency to build from source.
+  * Another possible mitigation is the [`--with-custom-python`](https://github.com/Homebrew/homebrew-core/pull/3548/files#diff-79aaf9c9ba469f17b9f16f996d7e6290R17) switch deployed for Mercurial, which has the desired effect but introduces class-level conditional dependencies, which don't behave well.
+* Homebrew uses one fewer system dependency.
+* ?
+
+
+## Alternatives
+
+* Do nothing: since the majority of Homebrew users use the most recently released version of OS X, Homebrew could elect to solve this issue by inaction if the forthcoming Sierra release linked the system python against a more recent TLS library. Unfortunately, the betas released to date continue to build the ssl module against OpenSSL 0.9.8.
```
Lukasa: A note here: Apple employees have *hinted* that future Python installs may link against LibreSSL, rather than OpenSSL, and macOS Sierra and El Cap do ship LibreSSL for other purposes (again without development headers). However, there is definitely no plan for Apple to do this in Sierra, and I wouldn't want to bet my life on it happening in 10.13 either.

Lukasa:
I'm a strong :+1:.

tdsmith:
What are the criteria for accepting proposals in this repository?

MikeMcQuaid:
@tdsmith I'm gonna say at least one approval and now you have one.

XX-avoid-system-python.md
tdsmith: I like this idea and I'll try to explore it this weekend.

MikeMcQuaid:
@tdsmith Definitely have enough to 🚢 this if you're happy. Nice work.

tdsmith:
I think we agree on the principle (🎉)  but I'm not sure about the best way
to implement it.

MikeMcQuaid:
I think moving to `depends_on "python"` in a few formulae without an option to disable and seeing if people :rage4: could be a good first step. Ansible feels like a nice test; it's widely used as an application rather than a library.

dch:
Thanks brewers for the effort you are putting in behind the scenes. As a drive-by contributor I'm reminded how much work it is to achieve things like the recent seamless switch from homebrew to homebrew-core, and working out all these dependency issues, on a platform where the default position from the manufacturer is to provide zero info.

MikeMcQuaid:
Thanks for the kind words @dch 💖

MikeMcQuaid:
@tdsmith :shipit:?

tdsmith:
Yes, lets. I'd like to proceed by adding `depends_on :python`.

tdsmith:
I will assign proposal number 6.

claco:
Apologies, maybe my questions were better placed here, but it was no longer 'open': https://github.com/Homebrew/homebrew-core/pull/5042#issuecomment-257978520

CamJN:
Eeeeeeewwwwww. Please don't do this. 😭

MikeMcQuaid:
@CamJN That's not a helpful comment and it's too late now.

CamJN:
How's this for helpful, these pages need to be updated as they are now wrong:

https://github.com/Homebrew/brew/blob/master/docs/Homebrew-and-Python.md
https://github.com/Homebrew/brew/blob/master/docs/Maintainer-Guidelines.md
https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md

The last one is only included because you've definitely moved to a pro-duplication strategy with the `@` formulae and now this.

MikeMcQuaid:
> How's this for helpful

It's not. Pull requests are welcome if you care to point out specifics that need correction beyond "they are now wrong".

> The last one is only included because you've definitely moved to a pro-duplication strategy with the @ formulae and now this.

No, we haven't.

PR #15, opened 2016-09-21
Hint at new location of migrated formulae

zmwangx:
Please read [the document](https://github.com/zmwangx/brew-evolution/blob/hint-at-new-location-of-migrated-formulae/015-hint-at-new-location-of-migrated-formulae.md).


015-hint-at-new-location-of-migrated-formulae.md
```diff
@@ -0,0 +1,71 @@
+# Hint at new location of migrated formulae
+
+## Introduction
+
+Users are confused when trying to install formulae known to have existed, yet
```
MikeMcQuaid: Can you point to some issues here?

015-hint-at-new-location-of-migrated-formulae.md
```diff
@@ -0,0 +1,71 @@
...
+getting the "No available formula" error due to said formulae having been
+migrated to another tap that has not been tapped. This proposal attempts to
+make the experience less confusing.
+
+## Motivation
+
+When a formula `foo` is migrated to another tap, subsequent attempts to install
+it produce the unhelpful error
```
MikeMcQuaid: Does the `search` not kick in?

015-hint-at-new-location-of-migrated-formulae.md
```diff
@@ -0,0 +1,71 @@
...
+Error: No currently available formula with the name "foo".
+A "foo" formula was migrated from alice/tap to bob/pat. You may want to try
+  brew info bob/pat/foo
+```
+
+When multiple matches are found in `tap_migrations.json` of multiple taps,
+multiple hints may be printed.
+
+When a match indicates migration to `homebrew/boneyard`, we may want to print a
+special warning.
```
MikeMcQuaid: Agreed; don't tell people to `tap` it but instead note it was removed. Arguably we should avoid auto-tapping `boneyard` too.

015-hint-at-new-location-of-migrated-formulae.md
zmwangx: Hmm, it does, not sure why I didn't realize (probably because I only tried `info clang-omp` where I don't have boneyard tapped). I'll add a paragraph on why search is not good enough: API rate limit easily exceeded without `HOMEBREW_GITHUB_API_TOKEN`; `homebrew/boneyard` not searched AFAIK (which accounts for half of the tap migrations in core); slow for people with bad internet etc.

015-hint-at-new-location-of-migrated-formulae.md
zmwangx: Will do, but I only know one instance (maybe this kind of thing just don't get reported — people are confused for a while, then move on).

015-hint-at-new-location-of-migrated-formulae.md
```diff
@@ -0,0 +1,71 @@
...
+
+Suppose a formula `foo` was migrated from the `alice/tap` tap to the `bob/pat`
+tap and hence there is an entry `"foo": "bob/pat"` in
+`Library/Taps/alice/tap/tap_migrations.json`, the "No available formula" error
+message could instead be
+
+```
+Error: No currently available formula with the name "foo".
+A "foo" formula was migrated from alice/tap to bob/pat. You may want to try
+  brew info bob/pat/foo
```
zmwangx: Need to tap `bob/pat` here, `info` does not autotap.

015-hint-at-new-location-of-migrated-formulae.md
MikeMcQuaid: Cool, I see there's one below.

015-hint-at-new-location-of-migrated-formulae.md
MikeMcQuaid: 👍

zmwangx:
Addressed the comments.

015-hint-at-new-location-of-migrated-formulae.md
```diff
@@ -0,0 +1,86 @@
...
+`tap_migrations.json` of the former home of `foo`, so we can encourage the user
+to look for `foo` in its new home.
+
+Note that currently `install` (but not `info`) follows up the "No available
+formula" error with a search, but the experience is bad for several reasons:
+
+- `brew search` does not search in homebrew/boneyard, which accounts for half
+  of the migrations from core;
+- `brew search` is slow, especially for connections with high latency;
+- Without `HOMEBREW_GITHUB_API_TOKEN`, a user can search for three or four
```
MikeMcQuaid: We also use the HTTPS GitHub authentication credentials which should reduce the number of folks affected by this.

015-hint-at-new-location-of-migrated-formulae.md
```diff
@@ -0,0 +1,86 @@
...
+Suppose a formula `foo` was migrated from the `alice/tap` tap to the `bob/pat`
+tap and hence there is an entry `"foo": "bob/pat"` in
+`Library/Taps/alice/tap/tap_migrations.json`, the "No available formula" error
+message could instead be
+
+```
+Error: No currently available formula with the name "foo".
+A "foo" formula was migrated from alice/tap to bob/pat. You may want to try
+  brew tap bob/pat
+  brew info foo
```
MikeMcQuaid: I'd suggest this says whatever command they tried to run rather than info. If it's install you could suggest `brew install bob/pat/foo`

MikeMcQuaid:
This is looking good. A few final nits then 🚢

015-hint-at-new-location-of-migrated-formulae.md
zmwangx: Hmm what do you mean by "HTTPS GitHub authentication credentials"? Do we reach into system keychain somehow?

MikeMcQuaid: Yes. See `github.rb`.

zmwangx: Interesting, thanks for the pointer. Very handy.

zmwangx:
@MikeMcQuaid Pushed two commits to address the outstanding comments.

MikeMcQuaid:
@zmwangx Still one unaddressed comment here but otherwise 👍

015-hint-at-new-location-of-migrated-formulae.md
zmwangx: This is already addressed in 	f9585b4.

015-hint-at-new-location-of-migrated-formulae.md
```diff
@@ -0,0 +1,99 @@
...
+
+```
+  brew install bob/pat/foo
+```
+
+if the user was attempting an install. (However, we might want to be cautious
+not to recommend `install` outright if `bob/pat` is a third-party tap;
+inspection prior to installation might be a better idea in that situation,
+since sandboxing is not enabled by default for third party taps, and hence they
+could do unmeasurable damage if seized by a malicious actor).
```
DomT4: This is a little OTT, given Homebrew doesn't regularly use `sudo` and unless someone comes up with a decent solution in the next 30 days or so won't be able to use `sudo`, so the damage would be limited to things the user had permission to fiddle with. A formula doing `rm -rf /` without `sudo` would fail, for example.

I don't think we should hype up Homebrew as a way to destroy your system, because used as intended (and possibly as enforced soon) it isn't.

The temptation is just to set a date where the sandbox goes on for everything. We've broken third-party taps in the past in the name of security, MD5, SHA1, etc.

zmwangx: I will remove the warning since it doesn't quite belong here, but `rm -rf ~` is enough damage (especially if the attacker is smart enough to list `~` and avoid `Library`, hence clearing everything except `Library`).

DomT4: Indeed. I agree it would be an enormous pain and exceptionally annoying for a user, and is something we definitely want to try to guard against, but theoretically you should only be able to bork a user rather than the system, which is some small mercy.

I think a proper solution is to just break stuff, given enough warning. There's just case here, IMO.

zmwangx:
I have removed the warning about autotap. (Sorry for the delay — I thought I corrected that already.)

MikeMcQuaid:
This repository was deprecated but we'd accept a PR for this.

PR #17, opened 2016-10-07
xz compression proposal

UnixJunkie:


scpeters:
There is a [comment in dev-cmd/bottle.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bottle.rb#L226-L227) saying that gzip was chosen instead of bzip2 since it is faster to compress and decompress.

Here's a [benchmark comparing gzip, bzip2, and xz](https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/). xz has the best compression ratio, and is faster to decompress than bzip2, but it's compression time can be much greater (up to 5x) depending on the compression level used.

scpeters:
The conclusion section of that [benchmark article](https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/) is very informative to read:

"So which should you use? It’s going to come down to using the right tool for the job and the particular data set that you are working with.

If you are interactively compressing files on the fly then you may want to do this quickly with gzip -6 (default compression level) or xz -1, however if you’re configuring log rotation which will run automatically over night during a low resource usage period then it may be acceptable to use more CPU resources with xz -9 to save the greatest amount of space possible. For instance kernel.org compress the Linux kernel with xz, in this case spending extra time to compress the file well once makes sense when it will be downloaded and decompressed thousands of times resulting in bandwidth savings yet still decent decompression speeds.

Based on the results here, if you’re simply after being able to compress and decompress files as fast as possible with little regard to the compression ratio, then gzip is the tool for you. If you want a better compression ratio to save more disk space and are willing to spend extra processing time to get it then xz will be best to use. Although xz takes the longest to compress at higher compression levels, it has a fairly good decompression speed and compresses quite fast at lower levels. Bzip2 provides a good trade off between compression ratio and processing speed however it takes the longest to decompress so it may be a good option if the content that is being compressed will be infrequently decompressed."

zmwangx:
Network I/O is more of a bottleneck than CPU cycles in most cases. IMO the main selling point of gzip is its prevalence.

MikeMcQuaid:
> There is a comment in dev-cmd/bottle.rb saying that gzip was chosen instead of bzip2 since it is faster to compress and decompress.

Yes, this is the case (I decided this).

Unfortunately `xz` is a non-starter for us because it does not ship with macOS by default (unlike `gzip` or `bzip2`). `bzip2` is too slow to extract so: `gz` it is for now.

zmwangx:
> Unfortunately xz is a non-starter for us because it does not ship with macOS by default.

`/usr/bin/tar` supports `xz` compression and decompression, although there's no `/usr/bin/xz` binary. So saying macOS doesn't ship `xz` isn't fair.

MikeMcQuaid:
For this proposal to be accepted I'd need to see it fleshed out with answers to the following questions:

- Which version(s) of OS X/macOS support `xz` from `/usr/bin/tar`?
- What's the compression time/decompression time/file size for 5-10 of our most popular bottles?

We'd need to see significant improvements without significant regressions on all of these metrics to consider this.

scpeters:
My yosemite machine (10.10.5) has bsdtar 2.8.3:

~~~
$ /usr/bin/tar --version
bsdtar 2.8.3 - libarchive 2.8.3
~~~

It appears to support xz:

~~~
$ /usr/bin/tar --help | grep '\-J'
  -z, -j, -J, --lzma  Compress archive with gzip/bzip2/xz/lzma
~~~

javian:
If this page is correct https://opensource.apple.com/ then bsdtar 2.8.3 has been in place since 10.7 . 2.8.3 was released 15th of March 2010 and 10.7 came out in July 2011. xz compression has been supported in gnu tar sine march 2009.

MikeMcQuaid:
Given we want to at least notionally support back to 10.5 we'd need to have something which can handle both types. If someone does the benchmarks and shows an improvement I'd be open to using it for new bottles going forward.

zmwangx:
I can answer the benchmark question. See the table below with results on the top 30 formulae, especially the last row, "total". Benchmark script is here: https://github.com/zmwangx/brew-misc/blob/master/bottles-gzip-vs-xz/benchmark. More results in https://github.com/zmwangx/brew-misc/tree/master/bottles-gzip-vs-xz/results. Note that I generated my benchmarks on a mid-2015 15'' rMBP with 2.5GHz i7 and 500GB SSD. I imagine decompression performance should be closer on an HDD, but I didn't test that configuration.

Basically, xz archives are 25%–50% smaller most of the time. Especially noteworthy is the `mysql` case, where we get a 60% reduction that's worth ~60MB. That's ~10 seconds on a 50 Mbps connection (to Bintray that is, which doesn't necessarily saturate your full bandwidth), or ~1 minute on a 10 Mbps one, or even longer on shitty connections, you get the idea. Compression takes 10 times as long (but it doesn't matter to users, is nothing compared to compile compile time, and is offset by the number of uses you get out of one bottle build), while decompression is only twice as long, but isn't so scary to begin with even for big bottles (untarring a .tar.xz bottle of mysql takes ~5 seconds in a total, ~2 seconds more than that of .tar.gz).

In total, the size reduction is ~130MB across the thirty bottles, and decompression time is merely ~10 seconds longer.

```
Columns:
 1 - formula
 2 - .tar.gz size
 3 - .tar.xz size
 4 - size percentage (.tar.xz / .tar.gz)
 5 - .tar.gz tar user+sys
 6 - .tar.gz tar real
 7 - .tar.gz tar user+sys
 8 - .tar.gz tar real
 9 - tar time percentage (user+sys, .tar.xz / .tar.gz)
10 - .tar.gz untar user+sys
11 - .tar.gz untar real
12 - .tar.gz untar user+sys
13 - .tar.gz untar real
14 - untar time percentage (user+sys, .tar.xz / .tar.gz)
```

1           |  2         |  3         |  4     |  5      |  6      |  7       |  8       |  9       |  10    |  11    |  12     |  13     |  14   
-------------|-------------|-------------|---------|----------|----------|-----------|-----------|-----------|---------|---------|----------|----------|--------
openssl      |  3810133    |  2192724    |  57.5%  |  0.540   |  0.557   |  5.014    |  5.083    |  928.5%   |  0.200  |  0.214  |  0.340   |  0.387   |  170.0%
pkg-config   |  240634     |  186444     |  77.5%  |  0.033   |  0.034   |  0.186    |  0.188    |  563.6%   |  0.005  |  0.006  |  0.018   |  0.019   |  360.0%
sqlite       |  1415122    |  637384     |  45.0%  |  0.177   |  0.181   |  1.066    |  1.076    |  602.3%   |  0.019  |  0.019  |  0.053   |  0.054   |  278.9%
libpng       |  446951     |  293020     |  65.6%  |  0.060   |  0.062   |  0.437    |  0.444    |  728.3%   |  0.013  |  0.014  |  0.028   |  0.030   |  215.4%
node         |  12006319   |  8317292    |  69.3%  |  1.594   |  1.610   |  18.593   |  18.714   |  1166.4%  |  0.551  |  0.578  |  1.037   |  1.071   |  188.2%
readline     |  793661     |  419720     |  52.9%  |  0.108   |  0.111   |  0.726    |  0.735    |  672.2%   |  0.017  |  0.018  |  0.041   |  0.042   |  241.2%
freetype     |  826157     |  445812     |  54.0%  |  0.101   |  0.103   |  0.811    |  0.818    |  803.0%   |  0.018  |  0.019  |  0.044   |  0.046   |  244.4%
glib         |  6939005    |  3748500    |  54.0%  |  0.950   |  0.965   |  9.311    |  9.363    |  980.1%   |  0.137  |  0.141  |  0.343   |  0.352   |  250.4%
git          |  12240990   |  4159240    |  34.0%  |  1.507   |  1.519   |  12.560   |  12.616   |  833.4%   |  0.302  |  0.312  |  0.483   |  0.497   |  159.9%
python       |  12226733   |  8942484    |  73.1%  |  1.398   |  1.408   |  17.665   |  17.731   |  1263.6%  |  0.443  |  0.454  |  0.900   |  0.918   |  203.2%
pcre         |  1861532    |  776176     |  41.7%  |  0.270   |  0.276   |  2.065    |  2.094    |  764.8%   |  0.051  |  0.054  |  0.091   |  0.096   |  178.4%
fontconfig   |  1206603    |  913116     |  75.7%  |  0.118   |  0.120   |  1.123    |  1.136    |  951.7%   |  0.063  |  0.065  |  0.119   |  0.124   |  188.9%
gdbm         |  167915     |  125060     |  74.5%  |  0.027   |  0.029   |  0.177    |  0.182    |  655.6%   |  0.007  |  0.007  |  0.015   |  0.015   |  214.3%
gettext      |  8216213    |  6269580    |  76.3%  |  0.670   |  0.686   |  5.928    |  5.965    |  884.8%   |  0.265  |  0.275  |  0.517   |  0.530   |  195.1%
xz           |  460065     |  305908     |  66.5%  |  0.060   |  0.061   |  0.521    |  0.527    |  868.3%   |  0.018  |  0.018  |  0.034   |  0.035   |  188.9%
imagemagick  |  8884302    |  6768116    |  76.2%  |  0.802   |  0.809   |  7.743    |  7.811    |  965.5%   |  0.232  |  0.241  |  0.579   |  0.589   |  249.6%
libtiff      |  1071955    |  686188     |  64.0%  |  0.121   |  0.123   |  1.105    |  1.111    |  913.2%   |  0.039  |  0.041  |  0.077   |  0.078   |  197.4%
libyaml      |  103673     |  56768      |  54.8%  |  0.017   |  0.018   |  0.090    |  0.093    |  529.4%   |  0.004  |  0.005  |  0.007   |  0.008   |  175.0%
libtool      |  1032492    |  656460     |  63.6%  |  0.144   |  0.147   |  1.520    |  1.531    |  1055.6%  |  0.038  |  0.040  |  0.072   |  0.076   |  189.5%
cmake        |  10572381   |  4315652    |  40.8%  |  1.555   |  1.581   |  12.302   |  12.426   |  791.1%   |  0.341  |  0.362  |  0.567   |  0.596   |  166.3%
youtube-dl   |  1369267    |  1323764    |  96.7%  |  0.059   |  0.061   |  0.440    |  0.447    |  745.8%   |  0.011  |  0.012  |  0.080   |  0.082   |  727.3%
autoconf     |  892872     |  545692     |  61.1%  |  0.122   |  0.125   |  1.183    |  1.197    |  969.7%   |  0.021  |  0.023  |  0.050   |  0.052   |  238.1%
libksba      |  135467     |  114060     |  84.2%  |  0.022   |  0.023   |  0.129    |  0.133    |  586.4%   |  0.006  |  0.007  |  0.013   |  0.014   |  216.7%
jpeg         |  297688     |  165364     |  55.5%  |  0.040   |  0.042   |  0.222    |  0.227    |  555.0%   |  0.008  |  0.009  |  0.017   |  0.017   |  212.5%
python3      |  13857185   |  10116336   |  73.0%  |  1.617   |  1.629   |  21.365   |  21.409   |  1321.3%  |  0.469  |  0.481  |  0.970   |  0.986   |  206.8%
mysql        |  97365099   |  38906560   |  40.0%  |  12.338  |  12.454  |  109.692  |  110.396  |  889.1%   |  2.667  |  2.791  |  5.044   |  5.217   |  189.1%
wget         |  573053     |  453720     |  79.2%  |  0.080   |  0.083   |  0.977    |  1.321    |  1221.2%  |  0.016  |  0.017  |  0.056   |  0.058   |  350.0%
pango        |  1023151    |  626872     |  61.3%  |  0.229   |  0.234   |  2.915    |  4.068    |  1272.9%  |  0.043  |  0.045  |  0.101   |  0.104   |  234.9%
vim          |  8254253    |  5503492    |  66.7%  |  1.924   |  2.487   |  18.480   |  23.200   |  960.5%   |  0.391  |  0.453  |  0.905   |  1.169   |  231.5%
go           |  82104657   |  52857852   |  64.4%  |  11.735  |  11.771  |  120.178  |  120.798  |  1024.1%  |  1.574  |  1.628  |  4.708   |  4.854   |  299.1%
total        |  290395528  |  160829356  |  55.4%  |  38.418  |  39.309  |  374.524  |  382.840  |  974.9%   |  7.969  |  8.349  |  17.309  |  18.116  |  217.2%

MikeMcQuaid:
Thanks @zmwangx!

> In total, the size reduction is ~130MB across the thirty bottles, and decompression time is merely ~10 seconds longer.

What's the compression time difference? Additionally, what's the size reduction/decompression time/compression time differences in terms of %s?

> 10 - .tar.gz untar user+sys
> 12 - .tar.gz untar user+sys

Should some of these be `.tar.xz`?

zmwangx:
> What's the compression time difference?

As said, xz compression (default) is roughly ten times as long, but it is dwarfed by typical compilation time, and shouldn't be much concern given the number of downloads and decompressions we get out of a single bottle build.

Check the table for more details.

> Additionally, what's the size reduction/decompression time/compression time differences in terms of %s?

Not sure I understand. Anything missing in the table?

> Should some of these be .tar.xz?

Yeah, sorry, updated.

zmwangx:
Let me single out the last row for clarity:

```
total	290395528	160829356	55.4%	38.418	39.309	374.524	382.840	974.9%	7.969	8.349	17.309	18.116	217.2%
```

Column reference is in https://github.com/Homebrew/brew-evolution/pull/17#issuecomment-252432739.

MikeMcQuaid:
> Let me single out the last row for clarity:

Mind editing in the column reference into the column headers instead of doing 1-14? Would make it easier for others (and me!) to follow. This may help: http://stevecat.github.io/table-magic/

zmwangx:
The concern was that putting the long descriptions in would make the tsv/csv unreadable on GitHub (it looks like [this](https://gist.github.com/anonymous/0866d4b9c96de28bc3855af17df0100a)), but the HTML table version seems bearable. Updated.

MikeMcQuaid:
What's the compression/decompression time look like for something like Qt5? My worry is even MySQL has some really heavy compression times there that aren't trivial compared to the build time. Is there any way we can speed up compression/decompression by using some parallelised method or tool?

At the moment it looks like: what do you want to optimise for: download time or compression/decompression time? It doesn't appear to be a clear win overall. I personally favour optimising for decompression time as that's something that is fastest when we have e.g. a cached build and I like optimising the no-op case.

zmwangx:
> What's the compression/decompression time look like for something like Qt5?

Similar to MySQL. ~130s compression, ~5s depression. (The size reduction is 92M -> 64M, ~30%.)

> My worry is even MySQL has some really heavy compression times there that aren't trivial compared to the build time. 

Depending on how you define "trivial". A MySQL build locally with all eight virtual cores took ~8 min. I estimate single core build to be ~50 min on my CPU. Meanwhile, single core compression takes ~2 min as shown in the table. It's not trivial, but not a bottleneck either.

I don't think you can get this out of `tar(1)` though.

> Is there any way we can speed up compression/decompression by using some parallelised method or tool?

Yes. `xz(1)` has the option `-T threads, --threads=threads`. Locally

```
$ time (xz -T2 qt5.tar)
( xz -T2 qt5.tar; )  129.79s user 0.77s system 196% cpu 1:06.42 total
$ time (xz -T4 qt5.tar)
( xz -T4 qt5.tar; )  139.54s user 0.81s system 369% cpu 38.007 total
$ time (xz -T8 qt5.tar)
( xz -T8 qt5.tar; )  191.18s user 1.43s system 612% cpu 31.444 total
```

so with 4 cores at least that does cut the compression time to 2–3 times that of gzip (~13s) instead of 10.

> At the moment it looks like: what do you want to optimise for: download time or compression/decompression time? It doesn't appear to be a clear win overall.

Optimizing for compression: that's a pretty weak point because the # downloads to # compressions ratio is easily >100 for a reasonably popular formula, and can go up to >10000 for something like MySQL.

Optimizing for download vs decompression: keep in mind that decompression is fast even for xz. For mysql it's just 5s with my CPU and SSD — 2s slower than gzip.

With a wimpy CPU it could be slower, but wimpy CPU are easy to upgrade and will be upgraded every few years, whereas network I/O is much harder to improve (requires overhaul on both ends).

The other factor is disk I/O. I didn't test decompression on HDD, but if disk I/O is the bottleneck on HDD, that will only make the gz vs xz difference irrelevant, because the difference lies in CPU cycles.

scpeters:
If you are concerned about compression time, you could try compression level 3 for xz, based on the shape of this plot from the [other comparison page](https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/).

plot

zmwangx:
`mysql` compression benchmarks:

```zsh
#!/bin/zsh
print 'level\tsize\tcompression time'
tar xf ~/Library/Caches/Homebrew/mysql-5.7.15.sierra.bottle.tar.gz
for level in {0..9}; do
    (
        TIMEFMT='%U %S'
        time (/usr/bin/tar -cJf mysql.tar.xz --options="compression-level=$level" mysql)
    ) |& sed 's/s//g; s/ /+/' | bc | read compression_time
    wc -c <mysql.tar.xz | read size
    printf '%d\t%s\t%.2f\n' $level $size $compression_time
    rm -f mysql.tar.xz
done
```

| level | size     | compression time |
|-------|----------|------------------|
| 0     | 74136416 | 18.80            |
| 1     | 65065072 | 24.00            |
| 2     | 60283884 | 31.53            |
| 3     | 46033604 | 37.72            |
| 4     | 44209688 | 68.08            |
| 5     | 40029424 | 89.18            |
| 6     | 39420208 | 106.25           |
| 7     | 39420208 | 106.24           |
| 8     | 39420208 | 107.57           |
| 9     | 39420208 | 114.26           |

bench

We should really optimize for users though.

zmwangx:
Throwing decompression time to the mix, which is more important:

| 0 | 74135524 | 18.11  | 7.14 |
|---|----------|--------|------|
| 1 | 65064808 | 20.94  | 6.30 |
| 2 | 60283560 | 27.84  | 5.93 |
| 3 | 46033276 | 33.22  | 5.04 |
| 4 | 44214748 | 65.99  | 5.02 |
| 5 | 40017904 | 88.91  | 4.96 |
| 6 | 39413300 | 107.72 | 4.80 |
| 7 | 39413300 | 106.25 | 4.68 |
| 8 | 39413300 | 107.21 | 4.69 |
| 9 | 39413300 | 108.89 | 4.92 |

out

scpeters:
If we value all three aspects, just looking at the shape of the curves, it seems like xz compression level 3 is a sweet spot, giving the bulk of improvements to bottle size and decompression time without drastically increasing the compression time.

If we multiply by the number of times each operation is likely to occur, though, compression time is the least important since it just happens once. As a developer, I'm often uninstalling and reinstalling bottles that are already downloaded, so I value quick decompression time, but I'm not sure how often most users reinstall formulae. @zmwangx asserted "99% of users download a bottle once and install it once", which seems like it could be true.

At risk of bike-shedding, is it worth looking at the analytics data to help inform this analysis? I don't know exactly what is in that data.

MikeMcQuaid:
> A MySQL build locally with all eight virtual cores took ~8 min. I estimate single core build to be ~50 min on my CPU. Meanwhile, single core compression takes ~2 min as shown in the table. It's not trivial, but not a bottleneck either.

Single vs. multiple core time doesn't matter unless we find a way of doing a multi-core compression of the tarball. As-is you're saying compression increases time by 25% of the MySQL build. The effects of that on our CI build times concern me.

> Optimizing for compression: that's a pretty weak point because the # downloads to # compressions ratio is easily >100 for a reasonably popular formula, and can go up to >10000 for something like MySQL.

As mentioned above: our CI is not an infinite resource so needs to be strongly factored it.

> Optimizing for download vs decompression: keep in mind that decompression is fast even for xz. For mysql it's just 5s with my CPU and SSD — 2s slower than gzip.

So an increase of 66%. 

> With a wimpy CPU it could be slower, but wimpy CPU are easy to upgrade and will be upgraded every few years, whereas network I/O is much harder to improve (requires overhaul on both ends).

Bottles exist to reduce CPU usage and world-wide CPU speeds are not increasing as fast as network I/O speeds. Network I/O can also be cached locally and organisationally.

> The other factor is disk I/O. I didn't test decompression on HDD, but if disk I/O is the bottleneck on HDD, that will only make the gz vs xz difference irrelevant, because the difference lies in CPU cycles.

The difference is not purely CPU cycles. When bottles were created a gziped tarball was faster to decompress than a non-gziped one.

> That doesn't benefit 99% of the users. 99% of users download a bottle once and install it once. Each on the CI the cached situation is pretty rare (except immediately after bottle creation).

That's speculative, you have no data to show that it's 99% so it harms your (otherwise well-researched) quantitive argument to produce numbers for this.

> Yes. xz(1) has the option -T threads, --threads=threads.

This seems like it would be a good option for bottling as long as it was not a hard dependency (i.e. required to create all bottles) and not a soft dependency for bottle consumers.

Relatedly, I think we require `xz` for decompressing `.tar.xz` bottles in Homebrew currently. If nothing else I'd ❤️ to see a patch come out of this that allows us to use a new enough BSD tar to extract `tar.xz` on new macOS versions. This would be useful regardless of what's decided here.

> If we value all three aspects, just looking at the shape of the curves, it seems like xz compression level 3 is a sweet spot, giving the bulk of improvements to bottle size and decompression time without drastically increasing the compression time.

I agree. On the other end, going above e.g. 7 has little benefit and 9 seems actively worse than 8. What's the default, out of interest?

> At risk of bike-shedding, is it worth looking at the analytics data to help inform this analysis? I don't know exactly what is in that data.

We'd need to collect additional data for that. I'm not opposed to that.

---

In summary, I'm somewhat convinced of the benefits now and for me it's more a discussion about what level of compression we should use rather than whether `tar.xz` is a winner or loser. I think we've reached enough a stage where this proposal (or a new one) could be improved and here's my suggested implementation plan:

- Add support for using `.tar.xz` source packages into Homebrew/brew with a new-enough BSD tar. This will validate that it behaves appropriately on these packages and remove a dependency for a bunch of people.
- Add support into Homebrew/brew for extracting `.tar.xz` bottles (possibly worth throwing in everything BSD tar supports in there)
- Add support to `brew bottle` (perhaps optional at first) to compress `.tar.xz` bottles and add the new DSL element required to use them.
- Get `brew test-bot` building, uploading and distributing `.tar.xz` bottles.

zmwangx:
> As-is you're saying compression increases time by 25% of the MySQL build.

Sure. 8 min => 10 min (assuming CI is actually that performant). We have many formulae that build much slower. Say rust. The compression time is almost always capped at 5 min though (and <<5 min in the vast majority of cases), because bottles only get so big. (The largest one that comes to mind is ghc, currently 195M, more than twice that of mysql. The single-threaded .tar.xz creation time is 5–6 min.)

> > Optimizing for download vs decompression: keep in mind that decompression is fast even for xz. For mysql it's just 5s with my CPU and SSD — 2s slower than gzip.
>
> So an increase of 66%.

That's not a fair evaluation. The bottleneck of pouring bottles lies in `Keg.relocate_text_files`. I just timed an `mysql` install. The untar part took some 2.5 seconds as expected. `relocate_text_files` took a jarring ~170 seconds. **Moving to a .tar.xz bottle only inflates the pouring time by 1–2% in this case.** This pattern can be seen in other formulae too, especially ones with a lot of text files where xz's compression power shines.

> This seems like it would be a good option for bottling as long as it was not a hard dependency (i.e. required to create all bottles) and not a soft dependency for bottle consumers.

It's easy to allow both gzip and xz bottles. As long as user's tar(1) supports xz, it doesn't matter whether it was created with threading.

> world-wide CPU speeds are not increasing as fast as network I/O speeds.

Evidence? CPUs today are mostly good enough (those that are not enough will be replaced soon anyway), while you can't say so for network speeds.

> Network I/O can also be cached locally and organisationally.

Assuming you're talking about LAN caching, that's only feasible for large organizations, or something like Travis. I don't know the percentage of such use cases. Also, I think we're optimizing for less privileged users in this proposal.

> The difference is not purely CPU cycles. When bottles were created a gziped tarball was faster to decompress than a non-gziped one.

Decompression speed is a CPU cycles problem. Decompressing an xz archive takes more CPU cycles, hence more time.

> That's speculative, you have no data to show that it's 99% so it harms your (otherwise well-researched) quantitive argument to produce numbers for this.

It would love to give a quantitative argument, but I don't think analytics data tell you that. Happy to be proven wrong. (One may try to compare GA and Bintray data, but remember, GA can be opted out.)

Without data, I would say the typical use cases are `brew install; brew update; brew upgrade`, which means each downloaded bottle is installed once. Of course the 99% figure is just symbolic.

> I agree. On the other end, going above e.g. 7 has little benefit and 9 seems actively worse than 8. What's the default, out of interest?

I believe the default level is 6.

> here's my suggested implementation plan

👍

MikeMcQuaid:
> That's not a fair evaluation. The bottleneck of pouring bottles lies in Keg.relocate_text_files. I just timed an mysql install. The untar part took some 2.5 seconds as expected. relocate_text_files took a jarring ~170 seconds. Moving to a .tar.xz bottle only inflates the pouring time by 1–2% in this case. This pattern can be seen in other formulae too, especially ones with a lot of text files where xz's compression power shines.

Good to know, I didn't realise that (because we didn't used to do that). Could you create a new issue about this? It's a pretty big deal and would love to see it improved.

> Evidence? CPUs today are mostly good enough (those that are not enough will be replaced soon anyway), while you can't say so for network speeds.

I have as much evidence for my claim as you have for the inverse 😉 

> I believe the default level is 6.

I'd be up for leaving it as the default.

zmwangx:
> Could you create a new issue about this?

Here or in brew?

> It's a pretty big deal and would love to see it improved.

Yeah, but all those placeholders need to be replaced somehow... Or just give up the placeholders.

MikeMcQuaid:
> Here or in brew?

Brew, cheers.

> Yeah, but all those placeholders need to be replaced somehow... Or just give up the placeholders.

Yeh, I think we're probably doing it currently in a hilariously ill-performant fashion that can be optimised (and I find optimising such things fun)

zmwangx:
https://github.com/Homebrew/brew/issues/1250.

UnixJunkie:
there are paralle implementations of xz. pxz for example is in brew. It allows to decompress in parallel for example.

zmwangx:
Re pxz:

1. pxz(1) is a parallel compressor, not a parallel decompressor. It decompresses with one thread only.
2. xz(1) already has the `-T, --threads` option, and performance is almost the same (the parallelization strategy is the same: split the file into parts and compress different parts concurrently).
3. You can't expect users to have pxz(1) installed.

In summary, pxz(1) is no better than stock xz(1).

UnixJunkie:
@zmwangx thanks for the details !

MikeMcQuaid:
> but you should add a new file instead of overwriting the proposal template. See #15 for an example.

@UnixJunkie Can you do this and flesh out the proposal a bit more? Thanks!

vitorgalvao:
Disclaimer: I have read only part of the discussion (a big part of it, about two weeks ago) and haven’t reached the final conclusions (stopped following in detail). I also have not yet read the article I’m abou to link to, but I found it while browsing and immediately though of this issue. This may also not be of interest to this discussion since it pertains to long-term archiving (however “long-term” is may be explained in the link), but with all the technical discussion in this post, even if irrelevant to the decision, it might be of interest to some of you.

“Xz format inadequate for long-term archiving”. [Original](http://www.nongnu.org/lzip/xz_inadequate.html) | [Hacker News discussion](https://news.ycombinator.com/item?id=12768425) (in good HN fashion, the top post starts with “This article is overblown in its criticisms and has some issues of its own”).

scpeters:
From the [top HN comment](https://news.ycombinator.com/item?id=12769379):

"5. The article is written by the author of lzip (which it compares favorably to xz), but does not disclose this. Admittedly it is hosted on the lzip site, but overall the article comes across as an opinionated hit-piece designed to sow doubt in a competitor."

So I'd take the article with a grain of salt.

zmwangx:
I read the linked article (skimmed the highly technical sections). Here are its key findings:

> Key findings include: (1) xz's extensibility is unreasonable and problematic; (2) xz is vulnerable to unprotected flags and length fields; (3) LZMA2 is unsafe and less efficient than the original LZMA; (4) xz includes useless features that increase the number of false positives for corruption; (5) xz shows inconsistent behavior with respect to trailing data; (6) error detection in xz is several times less accurate than in bzip2, gzip and lzip.

Also, from gloassary:

> Unsafe: 1 Likely to cause severe data loss even in case of the smallest corruption (a single bit flip). 2 Likely to produce false negatives.

As you can see, the focus is very very much on corruption, which might be a problem for long term archival (mostly if you are brave enough to only have one copy), but not really a concern for us, because we checksum everything. If a downloaded bottle is corrupted, download it again. If a bottle is corrupted on Bintray, which is highly unlikely, we build and upload it again. Bottles are reproducible (not in the bit-to-bit sense) by design.

Other non-corruption related design issues are mostly theoretical, I highly doubt they affect daily usage, as long as we stick to the standard feature set. Software projects (including big names like Emacs) have been distributing xz tarballs for years and people seem to be happy with it. I also question the validity of certain criticisms, because while the author of this article might be very smart and knowledgeable, I daresay the developers behind xz are no idiots either. The only concern that's slightly worrisome is

> The xz format lacks a version number field.

but I highly doubt xz developers will ever break backwards compatibility without a clear migration path.

MikeMcQuaid:
This repository was deprecated but we'd accept a PR for this.

PR #18, opened 2016-10-14
Wait for another brew process

maxnordlund:
This came up when I was doing some maintenance, and while running `brew update && brew upgrade`, I wanted to uninstall some packages. But fortunately brew stopped me, but then I thought, why not wait for that other command to finish and then run this one. Still safe, but more convenient.


zmwangx:
`brew update` is typically fast, but `brew` also blocks on e.g. `install` which could last for minutes to hours. Not sure waiting is the best idea. Also, how do you communicate with the user? An endless stream of dots? Granted, the user could ^C or ^Z in those cases.

Another case to keep in mind is scripting. Say you have a batch of related brew commands to execute, and most likely the script is written with a still background environment in mind. When another brew command actively changed the environment, you may want to fail outright and re-evaluate the situation, rather than wait for its actions to complete and continue on as if nothing happened.

maxnordlund:
Maybe with a nice warning, like:

```
Another active Homebrew process is already in progress.
Press Ctrl-C to abort, or wait until it has finished.
```

Yes, either a flag, which isn't the best UX, or detecting non-interactivity/TTY.

zmwangx:
Sounds good 👍

zmwangx:
Actually, there's still a bit of problem with

> detecting non-interactivity/TTY

because a script's stdout and stderr could very well be connected to tty.

maxnordlund:
Hm, then I don't know. There isn't any other way to check for interactivity? In fish you can ask the shell with `status --is-interactive`, maybe there something similar for bash?

zmwangx:
`brew` can't talk to its parent process. We know the PPID, that's about it.

maxnordlund:
There may be some hope https://www.gnu.org/software/bash/manual/html_node/Is-this-Shell-Interactive_003f.html

If nothing else, we can always play it safe and bail if we're not sure brew is run interactively.

zmwangx:
> There may be some hope https://www.gnu.org/software/bash/manual/html_node/Is-this-Shell-Interactive_003f.html

That's for checking the interactivity of *the current shell*, not the parent shell. That is, you can check the interactivity of `/bin/bash` (or `/bin/zsh`, or whatever) within that `/bin/bash` process, but not from a `brew` process launched from that shell.

maxnordlund:
```sh
maxno[0]~$ bash -c 'ruby -e "puts STDOUT.isatty"'
true
maxno[0]~$ bash -c 'ruby -e "puts STDOUT.isatty" | cat'
false
```

This seem to work, and I think I got the sub shell thing right.

zmwangx:
In the second case, `ruby`'s stdout is connected to the pipe, which isn't a tty.

But you can't tell whether `ruby` is run at the command line, or via a script like

```sh
#!/bin/sh
ruby -e 'puts STDOUT.isatty'
```

when the script is run from an interactive shell without redirection of standard streams.

maxnordlund:
Right, so it should be
```sh
maxno[0]Downloads$ bash -c 'ruby -e "puts STDOUT.isatty"' | cat
false
```

maxnordlund:
Hm, let me try a script, hold on.

maxnordlund:
```sh
maxno[0]Downloads$ bash test.sh
true
maxno[0]Downloads$ bash test.sh | cat
false
```

Using the script you wrote.

zmwangx:
Let's get this straight. You want to learn from within `brew` whether it is run from a script, whose standard streams could be connected to a tty. Think for a minute and you'll realize why none of what you proposed would work, and why I said

> brew can't talk to its parent process. We know the PPID, that's about it.

maxnordlund:
So, the brew command is can be run from either a script, or directly from a terminal. In the latter case it's easy to ask STDOUT.isatty? which also would handle pipes `brew <foo> | bar`

But if it's run inside a script, that is not piped itself as demonstrate above, then the question is: should it still assume interactivity or not. I think that it should, but that's just my opinon.

---

I've been playing around a bit more and running this script:

```sh
#!/bin/sh
echo $-
echo $PS1
bash -c 'echo $-'
test -t 0; echo $?
test -t 1; echo $?
ruby -e 'puts STDOUT.isatty'
```

gives

```
hB

hBc
0
0
true
```

So the `test` thinks it's interactive because stdin/out isn't redirected, but `$-` tells a different story. It doesn't include the `i` flag. But if called with `-i`, `bash -i test.sh` then you get:

```
himB
\s-\v\$
hBc
0
0
true
```

So I do think it's possible to detect if the brew entry script is run interactively or not. Or just abandoned trying to be smart and ask the user for a flag if they need to run it in a script, like `--fail-fast` or `--no-wait`.

http://www.tldp.org/LDP/abs/html/intandnonint.html

maxnordlund:
Hm, yeah you are right in that it's nigh impossible to determine if the command was directly invoked by the user, but that doesn't necessarily mean it's not interactive. Anyways, I need to sleep, it's 2:30 here. Maybe I can understand this better after that.

MikeMcQuaid:
To take a step back: if this is purely about using `update`'s `lock` then the better solution is "don't run `update` twice at once". Obviously in this case one of the updates may have been caused by an auto-update in which case I think the better fix would be to skip the `update` if it's locked and continue on with whatever else (as it's possible to do multiple installs at once).

maxnordlund:
No, or at least I didn't think of it like that. Basically, if some long running brew process currently holds the lock, and I want to run something else that needs the lock, then wait for the first process to exit/release the lock instead of bailing with an error message.

After thinking a bit about it, I think having it block by default is fine. To turn it off, just use either a flag or an environmental variable (like `DEBIAN_FRONTEND=noninteractive`). The only "smartness" would be to do a simple check for `STDOUT.isatty? && STDIN.isatty?`, and leave it at that as good enough.

Maybe that helps clear things up a bit, 😄

MikeMcQuaid:
> some long running brew process currently holds the lock

Which lock? This seems to be just talking about the update lock (which is all that Bash lock is used for currently).

maxnordlund:
I guess, if that's what's `lock.sh` is used for. But because install does an auto-update, it fails when there's another process that does the same thing. So I would like it to wait for the update to complete, and when the lock is released, continue.

In order for scripts to be able to circumvent this blocking behaviour, there needs to be a flag or environmental variable to turn it off. The only "smartness" that I think it can have would be to check if stdin is a TTY, which indicates that the user can ^C to cancel the block.

MikeMcQuaid:
> So I would like it to wait for the update to complete, and when the lock is released, continue.

That seems reasonable if it's scoped exclusively to the auto-update/update behaviour. In that case, though, I almost just wonder if a better error message would be appropriate where we tell people they don't need to manually run `brew update` any more due to auto-updates?

maxnordlund:
Hm, what about multiple install commands in different terminals? Won't that also trigger this, because each one tries to auto-update?

MikeMcQuaid:
Yep, that's a case worth handling if for this to be sufficiently scoped 👍

maxnordlund:
OK, so then I propose that the `lock.sh` script is adjusted like so: If the flag/env variable is set or stdin isn't a TTY, do as today and bail out if it can't acquire the lock. Otherwise poll it every X seconds until a lock is acquired, and the proceed as normal.

Maybe it should still bail after some long timeout, say 5 min, just in case it's about to hang. But on other hand, that might be bad if updates take a long time, e.g. because of slow internet connection.

If the process that acquires the lock writes its PID, then the waiting process might be able to use that first, instead of polling. But of course it still has to acquire the lock before proceeding.

MikeMcQuaid:
That seems all very reasonable 👍

MikeMcQuaid:
This repository was deprecated but we'd review a PR for this.

PR #19, opened 2016-10-20
more bottles

UnixJunkie:
Please use the [proposal template](/proposal_template.md) for your pull request.

### DIFF
--- a/proposal_template.md
+++ b/proposal_template.md
@@ -1,24 +1,18 @@
 # $FEATURE_NAME
 ## Introduction
-A short, single-paragraph "elevator pitch" describing what the feature is.
+We should have more bottles (binary packages).
+Too often, brew is compiling things.
+That's too slow.
 
 ## Motivation
-Describe the problems that this proposal seeks to address. If the problem is
-that some common pattern is currently hard to express, show how one can
-currently get a similar effect and describe its drawbacks. If it's completely
-new functionality that cannot be emulated, explain why this new functionality
-would help Homebrew users.
+Unix users are demanding.
+Don't make them wait.
 
 ## Proposed solution
-Describe a solution to the problem. Provide examples and describe how they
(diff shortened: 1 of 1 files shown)


proposal_template.md
```diff
@@ -1,24 +1,18 @@
...
-Describe a solution to the problem. Provide examples and describe how they
-work. Show how your solution is better than current workarounds: is it cleaner,
-safer, or more efficient?
+Create as many bottles as necessary and as possible.
 
 ## Detailed design
-Describe the design of the solution in detail. The detail in this section
-should be sufficient for someone who is *not* one of the authors of this
-proposal to be able to reasonably implement the feature.
+I dont' know.
```
vitorgalvao: Typo (should be “don’t”).

proposal_template.md
```diff
@@ -1,24 +1,18 @@
 # $FEATURE_NAME
```
vitorgalvao: You forgot the title.

vitorgalvao:
You’re not supposed to edit the `proposal_template.md` itself, but to make a copy of it for your proposal. As it stands, you would be removing the template for other people.

As for the proposal itself, you may want to revise it. Currently it is the same as saying “implement more features” without being specific about what those are or how they’d work. Adding bottles is on-going work. It’s done as the rhythm time and resources allow — see the [Bottles documentation](https://github.com/Homebrew/brew/blob/master/docs/Bottles.md) for details on how they’re created.

Simply asking for more bottles does not produce an effect, and is not a tangible improvement to the software itself (i.e. the kind of proposal adequate for this repository). A better approach would be to ask first “why aren’t there any more bottles?” and possibly even offering your help if the response is something you are able to help on.

MikeMcQuaid:
> Create as many bottles as necessary and as possible.

We do already.

PR #20, opened 2016-10-24
added 007-use-xz-compression-for-archives.md

UnixJunkie:


UnixJunkie:
related to https://github.com/Homebrew/brew-evolution/pull/17#issuecomment-255528527

zmwangx:
Please do this in the original proposal. We don't want to lose the whole discussion.

MikeMcQuaid:
> Please do this in the original proposal. We don't want to lose the whole discussion.

Agreed

PR #21, opened 2016-10-25
Add Node Requirement Support

pmyers88:
This proposal suggests adding support for Node Requirement (`depends_on
:node`) to the Formula DSL.

If this is something you want to do I'd happily submit a PR :).

### DIFF
--- a/XX-add-node-requirement-support.md
+++ b/XX-add-node-requirement-support.md
@@ -3,18 +3,17 @@
 Phil Myers, 10-26-2016
 
 ## Introduction
-This proposal suggests adding support for Node Requirements (`depends_on :node`) to the Formula DSL.
+This proposal suggests adding support for Node Requirement (`depends_on :node`) to the Formula DSL.
 
 ## Motivation
-There are many ways to install `node`. Implementing this proposal would allow users who installed `node` without using `brew` to not have to unnecessarily re-install `node` using `brew` for `node` dependencies.
+The Formula DSL currently requires the `node` dependency to be installed via `brew`. There are many ways to install `node`. Implementing this proposal would allow users who installed `node` without using `brew` to not have to unnecessarily re-install `node` using `brew` for formulas with `node` dependencies.
 
-An issue that would be solved by adding support for `depends_on :node` exists [here](https://github.com/Homebrew/homebrew-core/issues/6187). I had the same problem. @MikeMcQuaid suggested this solution.
+An issue that would be solved by adding support for `depends_on :node` exists [here](https://github.com/Homebrew/homebrew-core/issues/6187). I had the same problem. Mike McQuaid suggested this solution.
 
 ## Proposed solution
 * Create a `NodeRequirement` class
 * Add `depends_on :node` to the Formula DSL
 * Update `Node-for-Formula-Authors.md` to suggest using `depends_on :node` instead of `depends_on "node"`
-* Update formulae? (maybe authors have to do this)
 
 ## Detailed design
 * Use any of the files in `brew/Library/Homebrew/requirements` as a template for writing the `NodeRequirement` class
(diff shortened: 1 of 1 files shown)


zmwangx:
Node installed from nvm is ephemeral (won't be able to available in another shell session without nvm loaded) and can vary wildly in version. Given the size and viralness of the Node.js community, many users probably don't understand their tooling well, and a lot of pointless support burden would arise due to this. So 👎 on this.

zmwangx:
Brewed node is only 40M in size so I don't understand the concern.

pmyers88:
The concern is that I prefer using nvm to manage my node versions and don't want to have an additional brewed node on my machine.

The Ruby community is comparable in size, viralness and misunderstanding of tooling, but brew doesn't force them to install brewed ruby versions unnecessarily. Why the difference?

zmwangx:
The profound difference is there's `/usr/bin/ruby` but not `/usr/bin/node`.

zmwangx:
Say we allow this `NodeRequirement`. Then the same can be requested for any language runtime (and more justifiably so for certain languages), and by extension any package that can be obtained via other means (read: all of them). The entire premise of brew breaks down.

pmyers88:
The `RubyRequirement` code doesn't make a distinction for `/usr/bin/ruby`, it just looks for a matching ruby version on the `$PATH`.

I'm still having trouble understanding why it's bad to check the `$PATH` for node before installing brewed node, but it's good to check the `$PATH` for ruby before installing brewed ruby. If the user has a misconfigured path they're still going to get the brewed executable; I'm not sure of the downside to checking.

I don't think that adding another language runtime requirement to a list of existing language runtime requirements leads to a slippery slope of destroying the brew ecosystem.

I do understand the hesitancy to add a new feature, particularly one that's not critical/in high demand. I just wanted to create this PR since it's something @MikeMcQuaid mentioned in [#6187](https://github.com/Homebrew/homebrew-core/issues/6187) and it's something I would like to see.

zmwangx:
In the case of Ruby or Perl, there's a `/usr/bin/` binary to fall back to, and installing the Homebrew formula (in `/usr/local` by default) would mask the system version, so there's a case to be made that users may want to use the system version. The requirements are written in a way that any binary on the path can satisfy the requirements, which is kind of unfortunate IMO, but at least you have something to fall back to in most cases.

In the case of Node, or anything not provided by macOS, it's very easy to have a different environment at run time than install time, and with shell-integrated language runtime version managers, it's especially easy to screw it up. A user may have nvm active when he was installing foo, then not have it later when he tries to run it, thus not having a runtime on PATH. Then, without sufficient knowledge of PATH (occurs surprisingly often), said user claims that it's Homebrew's fault.

If you really dislike the additional 40MB, `brew unlink node; rm -rf /usr/local/Cellar/node`, which should work even when we start to forbid uninstalling active dependencies. Problem solved.

zmwangx:
In short, requirements should be reserved for situations when something is conditionally fulfilled by the system (ruby, perl, emacs, etc., where the system version may not be enough), or when something isn't easily fulfillable within Homebrew (java, osxfuse, tex, tuntap, x11, etc., where we need a binary installer), not for "I can install X from somewhere else so can I skip X".

dunn:
FWIW some previous discussion at
- https://github.com/Homebrew/legacy-homebrew/pull/38789; and 
- https://github.com/Homebrew/legacy-homebrew/issues/44819

pmyers88:
Understood. Thanks for the input!

MikeMcQuaid:
I'm for this but will give it a wee while to see if other maintainers agree/disagree.

MikeMcQuaid:
Note something that may or may not cause this to be useless: anything that's bottled and has a dependency on `node` will force the usage of the `node` formula as the build-time and run-time versions are not trivially interchangeable for requirements/dependencies.

DomT4:
From my past experience on this, I think a special dependency here will end up causing Homebrew hassle personally, pretty much as neatly summarised by @zmwangx's comments.

I note that I actually proposed this multiple times in the past, particularly when the whole iojs situation went down, and I think time and Homebrew's Node stability has proven other maintainers here at the time who swatted away those PRs correct.

MikeMcQuaid:
This repository was deprecated but we're reviewing a PR for this.

PR #22, opened 2016-10-30
Add notice for configurable Auto-Update option

joshdekock:
Related to Homebrew/brew#288.


MikeMcQuaid:
This is documented in the manpage and we don't want to encourage people to disable it as it's useful for us in decreasing maintenance.

joshdekock:
It's a breaking change though, it's not backwards compatible. It's really not obvious that it's an option at all. Sure, I looked in the manpage, but most users won't use manages. Not mentioning it all all is silly. 

Even something which advises not to would be 1000x better, like: `Auto-update has been enabled, you can HOMEBREW_NO_AUTO_UPDATE=true in your shell profile to turn it off. This is not recommended though as <insert reason here>`. (I'm not how it decreases maintenance, but I'm sure you know, so you can insert it there.)

MikeMcQuaid:
> It's a breaking change though

It doesn't break anything.

> It's really not obvious that it's an option at all.

The same goes with most of Homebrew's options. This is somewhat intentional: it's designed such that the vast majority of users will not and should not set them.

bfontaine:
> I'm not how it decreases maintenance

It ensures Homebrew is always up-to-date and saves us (and users) from tons of issues where people install outdated formulae and then complain they’re broken.

zmwangx:
https://github.com/Homebrew/brew/pull/1413 may help a little bit.

PR #23, opened 2016-11-04
Add header 'Accept: application/octet-stream' to download requests

stefansundin:
Here is a proof of concept: https://github.com/stefansundin/brew/commit/75154c358fc55cbe82892b15aace792a36b91061


023-download-via-github-api.md
```diff
@@ -0,0 +1,38 @@
...
+Once you specify that header, you can have a formula with an url like this:
+```
+  url "https://api.github.com/repos/stefansundin/private-tool/releases/assets/1234567?access_token=#{ENV["HOMEBREW_GITHUB_API_TOKEN"]}"
+```
+
+You need to look up the asset id after uploading the file. This can be done by
+navigating to: https://api.github.com/repos/stefansundin/private-tool/releases?access_token=GITHUB_TOKEN
+
+## Proposed solution
+Add the header required to all downloads.
```
MikeMcQuaid: I think a better solution is allowing this to be optionally added to only api.github.com downloads.

023-download-via-github-api.md
```diff
@@ -0,0 +1,38 @@
...
+requires that you specify an extra header
+([as described here](https://developer.github.com/v3/repos/releases/#get-a-single-release-asset)).
+
+Once you specify that header, you can have a formula with an url like this:
+```
+  url "https://api.github.com/repos/stefansundin/private-tool/releases/assets/1234567?access_token=#{ENV["HOMEBREW_GITHUB_API_TOKEN"]}"
+```
+
+You need to look up the asset id after uploading the file. This can be done by
+navigating to: https://api.github.com/repos/stefansundin/private-tool/releases?access_token=GITHUB_TOKEN
```
MikeMcQuaid: How would the access token be set in a formula?

stefansundin: It could be set like above (with HOMEBREW_GITHUB_API_TOKEN). But it could vary and be determined by the formula author.

Looking up the asset id is a one time thing that the formula author has to do. It doesn't have to be automated.

MikeMcQuaid: For us to implement this I'd like to have this figured/fleshed out further.

stefansundin:
In case anyone finds this proposal, here's a formula that works without any changes to Homebrew itself:

You need to do this:
- Look up the asset id from the API, using an url like this: https://api.github.com/repos/stefansundin/private-tool/releases?access_token=GITHUB_TOKEN
- Ask your users to create a GitHub token with the "repo" scope and set the environment variable GITHUB_TOKEN.

```ruby
require "net/http"

class PrivateTool < Formula
  desc "Download a binary from a private GitHub repo."
  homepage "https://github.com/stefansundin/private-tool"
  version "1.0.0"
  sha256 "0f119da204025da7808273fab42ed8e030cafb5c7ea4e1deda4e75f066f528fb"

  api_url = "https://api.github.com/repos/stefansundin/private-tool/releases/assets/1234567?access_token=#{ENV["GITHUB_TOKEN"]}"
  uri = URI(api_url)
  http = Net::HTTP.new(uri.host, uri.port)
  http.use_ssl = true
  response = http.request_head(uri.request_uri, {"Accept" => "application/octet-stream"})
  if response.code == "302"
    url response["location"]
  else
    onoe "Unable to resolve GitHub download link. Please check if $GITHUB_TOKEN is set."
    onoe "Test this url: #{api_url}"
  end

  def install
    bin.install "private-tool"
  end
end
```

023-download-via-github-api.md
```diff
@@ -0,0 +1,38 @@
...
+
+There is a chance that this would break other downloads, but I think this could
+be tested.
+
+## Detailed design
+I made a proof of concept that I confirmed works: https://github.com/stefansundin/brew/commit/75154c358fc55cbe82892b15aace792a36b91061
+
+## Alternatives considered
+Adding a `private true` flag that looks up the download url via the GitHub API
+automatically, perhaps using the `HOMEBREW_GITHUB_API_TOKEN` variable for this.
```
tdsmith: Can you comment on why this approach wasn't chosen?

stefansundin: I am not familiar enough with the Homebrew source code to code this solution.

MikeMcQuaid: We can help you with such things.

MikeMcQuaid:
This repository was deprecated but we'd review a PR for this.

PR #24, opened 2016-11-10
Propose Python Application Runtime

tdsmith:


XXX-python-application-runtime.md
```diff
@@ -0,0 +1,68 @@
+# Homebrew Python application runtime
+
+## Introduction
+Package a "secret" keg-only Python distribution to use as a runtime for Python applications in order to reduce the side effects associated with `brew install`ing a Python application.
+
+## Motivation
+
+* HEP 6 described the motivation for avoiding system Python for applications installed by Homebrew and recommended that applications are packaged with `depends_on :python`.
```
scpeters: I didn't know what HEP meant, but I guess it means "Homebrew Evolution Proposal"? This is the first time I've seen the acronym.

XXX-python-application-runtime.md
```diff
@@ -0,0 +1,68 @@
...
+
+* never uses system Python, and
+* never changes the system's default python interpreter,
+
+and to amend HEP 6 to use this new method.
+
+
+## Proposed solution
+
+The proposed solution is to add two additional keg-only formulas to homebrew-core, packaging CPython 2.7 and CPython 3. Instead of using `depends_on :python(3)`, all formulas that install Python applications will depend unconditionally on one of these new, keg-only formulas.
```
MikeMcQuaid: I don't think Python 3 is necessary given it's not provided with macOS.

XXX-python-application-runtime.md
```diff
@@ -0,0 +1,68 @@
...
+Some desired attributes of the design are:
+
+* The Homebrew Python application runtime (PAR) will be completely independent from the `python` or `python3` formulæ and any other installed Python distributions. Installing the PAR must not change the behavior of any other installed Python distribution. Installing the PAR must not change the default system Python distribution.
+* The PAR may never be linked into HOMEBREW_PREFIX.
+* Formulas may not install packages to the PAR's prefix.
+* The PAR will ship with the virtualenv package, which will subsume the function of the virtualenv `resource` on the `Homebrew::Language::Python::Virtualenv` mixin module.
+
+A present deficiency of this proposal is the absence of a nifty beer metaphor.
+
+A technical concern is the duplication of the Python environment on disk. The PAR will consume ~56 MB. This is considered to be acceptable.
```
MikeMcQuaid: Are there parts we can delete if it's just an application runtime? Could we consider making the `python` formula share some stuff with this?

XXX-python-application-runtime.md
```diff
@@ -0,0 +1,68 @@
...
+
+Having multiple Python environments installed is often considered undesirable. However, the PAR distributions will be hidden from the user's environment and will not affect the behavior of any other Python environment, so this is unlikely to contribute to user confusion. A private PAR represents an improvement relative to a dependency on the user-visible Homebrew python.
+
+Another technical concern is that the Apple system python has non-standard behavior with respect to trust stores and the system Keychain; this change will render it impossible to install a Homebrew application which takes advantage of this behavior. This deficiency is considered acceptable because it is not feasible to adopt this behavior in the PAR since it relies on patches to CPython which will not be adopted upstream, and the behavior changes are of niche interest. Further, the trust store behavior is most interesting to corporate environments capable of managing application deployments without Homebrew.
```
MikeMcQuaid: I think we should apply the Apple patches. I don't think matching Apple's Python or the trust store are of niche interest.

XXX-python-application-runtime.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Alternatives considered
+
+* Make the `python` formula keg-only.
```
MikeMcQuaid: I still probably favour this.

XXX-python-application-runtime.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+This is not preferred because a) it represents a breaking change and b) the user experience for keg-only formulas can be poor.
```
MikeMcQuaid: We should consider making the user experience better. I think manually adding something to your PATH is better than maintaining an identical package to avoid it being added to your PATH.

XXX-python-application-runtime.md
tdsmith: Yes, that's my intended use.

XXX-python-application-runtime.md
tdsmith: It is simpler not to; 60 MB is small enough that I don't perceive a good return on investigating this further. If anyone's interested in doing this work, we can always accept PRs after the formula lands.

XXX-python-application-runtime.md
MikeMcQuaid: I'd rather this was scoped just to Python 2 rather than creating another formula people may not use. I've never seen people complaining about e.g. our Python 3 being used over a pyenv Python 3.

XXX-python-application-runtime.md
tdsmith: My colleagues in the lab don't know what a PATH is and I would be happier if I can help them set up a scientific Python environment to run my analysis package without having to teach them how to edit their shell profile. :)

There are opportunities for improving keg management (imagining some sort of `brew keg attach` environment...) but I think that is a longer-term goal and I really don't want to complicate `brew install python` for beginning developers.

Anyway, accepting this maintenance burden to simplify the cognitive load on users seems like a good tradeoff to me.

XXX-python-application-runtime.md
tdsmith: The fundamental issues are the same. We do have applications that use Python 3, the Python ecosystem is finally pivoting away from Python 2 as it creeps towards EOL, it's reasonable for a user to want to continue using 3.5 as their system python3 for development after the python3 formula switches to 3.6 and pyenv is how we recommend users do that and it would be polite not to step on their environments.

If it's implemented, there are definitely formulas which will use it right away.

MikeMcQuaid: I think it's fine for us to stomp on a `pyenv` as that requires manual PATH setup. I think we should just `depends_on "python3"` and not open it up to all the custom nonsense that's plagued Python 2. No other package managers let you do this!

MikeMcQuaid: Regardless: let's focus on and solve Python 2's issues for now and we can handle Python 3 later.

tdsmith: I can live with that, I guess, but it seems odd to have good isolation for Python 2 apps and not for Python 3 apps. It's the same problem, solved the same way.

XXX-python-application-runtime.md
MikeMcQuaid: I'm just fed up with the special snowflake treatment of both Python users and the ecosystem. We wouldn't consider this approach for anything else and this is exactly the reason most of these things used to be keg_only. I think with an announcement date of when we're going to do this with `python`'s formula which includes instructions on how to set your PATH is sensible. Similarly, I think we hold off migrating any more package across until then.

Alternatively, I think we back out on https://github.com/Homebrew/brew-evolution/pull/12 entirely and don't let formulae depend on Python and go back to using system Python for everything and adding `with-custom-python` where necessary (e.g. `vim`).

XXX-python-application-runtime.md
MikeMcQuaid: It's not the same problem because there's no system Python 3.

XXX-python-application-runtime.md
zmwangx: > Alternatively, I think we back out on #12 entirely and don't let formulae depend on Python and go back to using system Python for everything.

Don't like the sound of that because IMO system Python will only get worse over time.

MikeMcQuaid: It won't get worse, it'll just get comparatively less featured.

XXX-python-application-runtime.md
MikeMcQuaid: I agree. `pyenv` is not affected in this case because the shims are added to the start of your PATH for it to function. Regardless, we're agreed on focusing on Python 2.

XXX-python-application-runtime.md
tdsmith: Just to be specific: system Python will not be able to make secure connections with sites that discontinue TLS 1.0. Sunset dates for TLS 1.0 are mandated by PCI and have been announced by CDNs. Call that what you want. ;)

We would also re-adopt the risk associated with Apple making a decision to discontinue providing a Python interpreter.

I strongly favor a) either adopting this proposal or b) taking no action over c) making Python keg-only or d) moving to repeal HEP 6.

MikeMcQuaid: What about if we had a keg-only mode that ran `brew link` if you did `brew install python` but didn't if it was pulled in as a dependency?

tdsmith: My first reaction is that increasing the number of keg states feels complex. It would be an excuse to add logic for remembering whether a package was explicitly installed or not, which I think people have been asking about for a while.

MikeMcQuaid: > It would be an excuse to add logic for remembering whether a package was explicitly installed or not, which I think people have been asking about for a while.

This was my thinking too.

ilovezfs:
>Alternatively, I think we back out on #12 entirely and don't let formulae depend on Python and go back to using system Python for everything and adding with-custom-python where necessary (e.g. vim).

That would be my preference.

I think #12 was a mistake, and #24 makes things worse and even more complicated than they already are.

MikeMcQuaid:
Have thought about this more and talked to @ilovezfs and these are my preferences:
- Use `depends_on "python"` for specific applications that have a primary use using TLS over a network (e.g. `ansible`)
  - People who care about the `$PATH` situation should manually use `/usr/bin/python` in their scripts
- Applications that don't have a primary use of using TLS over a network should stay using the system 2.7 Python (and `depends_on :python if MacOS.version <= :snow_leopard` or no `depends_on` at all)
- Formulae that have a specific, obvious, requested need for a custom Python (e.g. `vim`) should have a `with-custom-python` option added. This should not be added at a global level.

ilovezfs:
Yes, I agree with the above. I'd add that my preference is that the `depends_on "python"` be used sparingly. We can always add it later to a given formula.

MikeMcQuaid:
@tdsmith Sure. Can you ask specific questions or state which parts you'd like elaboration on, please?

tdsmith:
I'm also surprised and disappointed that we're considering rejecting HEP 6.
I thought I did a pretty good job of describing the impacts and
alternatives and trying to invite discussion and suddenly that feels like
wasted effort.

On Fri, Nov 11, 2016, 07:39 Tim D. Smith <tim.smith@uci.edu> wrote:

I don't think statements to take positions are very helpful in these
discussions without an explanation of the technical and design
considerations that have led you to take them. Can you (both) elaborate?

On Fri, Nov 11, 2016, 00:54 ilovezfs <notifications@github.com> wrote:

Yes, I agree with the above. I'd add that my preference is that the depends_on
"python" be used sparingly. We can always add it later to a given formula.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Homebrew/brew-evolution/pull/24#issuecomment-259911031>,
or mute the thread

MikeMcQuaid:
> I'm also surprised and disappointed that we're considering rejecting HEP 6.
> I thought I did a pretty good job of describing the impacts and
> alternatives and trying to invite discussion and suddenly that feels like
> wasted effort.

I don't see it as rejecting but not adopting it for applications that do not require it. I think it's inevitable that we're going to see iteration on these things after they are rolled out to users.

> Use `depends_on "python"` instead

I said I was 👍 on this in the original proposal multiple times and you decided to go ahead with `depends_on :python` anyway. That was up to you but I don't think it's unreasonable for me to restate my position on that here.

MikeMcQuaid:
@tdsmith Any more news here?

MikeMcQuaid:
This repository was deprecated so closing this out.

